PR #20, opened 2018-02-27
report button

jjc387:


VinishaM:
the div that holds review text is too wide - can we restore the padding/margins currently in the dev branch?

PR #28, opened 2018-04-07
Change most popular classes to most reviewed subjecs

julian-londono:
Changed the popular classes component to show a numbered list of the most reviewed subjects.


VinishaM:
- the Subjects collection contains a mapping from subject abbreviation to the full subject - can this be used instead of a separate mapping?
- rename topClasses method, rename imports/ui/PopularClasses.jsx, and update code comments to reflect what these files/functions now do
- Use <ol></ol> tags if possible to make an ordered list in PopularClasses list

PR #92, opened 2019-09-28
Setup Linter and CI

SamChou19815:
#### Linter

Linting is very broken now so I have to suppress almost everything in .eslintignore. The fact that we currently pass the linter test does not mean we are in a good state. We should eventually remove everything in .eslintignore and make it still pass linter!

The code style is not consistent in the repo. We should fix the linter problems first, then enforce a consistent code style!

#### CI

I only make the CI run linter because I can't figure out how to build the site.

### Test Plan

See the CI run!

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,5 @@
+imports/api/*.js
+imports/ui/js/
+imports/ui/*.jsx
+imports/ui/*.js
+server/
--- a/package.json
+++ b/package.json
@@ -1,8 +1,10 @@
 {
   "name": "course-reviews",
   "private": true,
+  "license": "UNLICENSED",
   "scripts": {
-    "start": "meteor run"
+    "start": "meteor run",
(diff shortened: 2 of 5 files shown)


.eslintignore
```diff
@@ -0,0 +1,5 @@
+imports/api/*.js
+imports/ui/js/
+imports/ui/*.jsx
+imports/ui/*.js
+server/
```
SamChou19815: These are not suppose to be ignored! They are ignored right now to make the linter CI job passing! They should eventually be removed from this file.

package.json
```diff
@@ -1,8 +1,10 @@
 {
   "name": "course-reviews",
   "private": true,
+  "license": "UNLICENSED",
   "scripts": {
-    "start": "meteor run"
+    "start": "meteor run",
+    "lint": "eslint . --ext .js --ext .jsx"
```
SamChou19815: Added to npm script so that everyone can run it very often

PR #99, opened 2019-10-06
Code cleanup: Fixed (most) ES Lint issues

drayfarley:
### Summary 

Fixed issues noted by ES Lint. Most changes were adding props validation in .jsx files.
In addition, the outdated use of React.findDOMNode was replaced by the proper convention


server/methods.js
```diff
@@ -523,10 +524,10 @@ function defaultDict() {
...
-  }
-}
\ No newline at end of file
+// function isJSON(str) {
+//   try {
+//     return (JSON.parse(str) && !!str);
+//   } catch (e) {
+//     return false;
+//   }
+// }
```
SamChou19815: Nits: Remove? Is it still useful?

server/publications.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable meteor/audit-argument-checks */
```
SamChou19815: 😮

imports/ui/Form.jsx
```diff
@@ -169,10 +180,10 @@ export default class Form extends Component {
...
       if (error || result === 1) {
         // Success, so reset form
-        ReactDOM.findDOMNode(this.refs.ratingSlider).value = 3;
-        ReactDOM.findDOMNode(this.refs.diffSlider).value = 3;
-        ReactDOM.findDOMNode(this.refs.workloadSlider).value = 3;
-        ReactDOM.findDOMNode(this.refs.profSelect).value = "none";
+        this.ratingSlider.current.value = 3;
+        this.diffSlider.current.value = 3;
+        this.workloadSlider.current.value = 3;
+        this.profSelect.current.value = "none";
```
SamChou19815: This is a good first step, but I still believe these types of imperative DOM manipulation can be eliminated. See https://reactjs.org/docs/refs-and-the-dom.html

julian-londono: That's true, I believe we can take advantage of reactive variables like state variables but good start on this.

julian-londono:
Creating a review app to check out before merging, will update link in a bit.

julian-londono:
Deployed at: https://cornellreviews-dev-pr-99.herokuapp.com/

server/publications.js
drayfarley: Basically that checks for if the arguments for Meteor methods are "checked" but the docs said you should only care if you're using the Meteor checks package, which we aren't.

https://github.com/dferber90/eslint-plugin-meteor/blob/master/docs/rules/audit-argument-checks.md

SamChou19815: @draycfarley Then it might be more helpful to disable it in `.eslintrc.json`>

julian-londono: Moved it to eslintrc in latest commit @SamChou19815

PR #101, opened 2019-10-09
Login redirect

julian-londono:
### Summary 

Google Login had become broken on Safari after enforcing HTTPS.  This was because Safari blocks the log-in pop-up by default.  In this commit, I have re-arranged the log-in workflow to use a series of redirects instead, allowing it to work with pop-up blockers and be more mobile-friendly.

### Test Plan 

Tested locally and will be testing in review app.

### Notes 

TODO:

This breaks adding users to the database when they update a review.  Very specific problem that will look to fix soon.  Essentially, the problem now is that since we redirect instead of use a callback function, we are not getting user's first and last name back from Google for some reason.

### DIFF
--- a/client/main.jsx
+++ b/client/main.jsx
@@ -2,9 +2,10 @@ import React from 'react';
 import { Meteor } from 'meteor/meteor';
 import { render } from 'react-dom';
 
-import Login      from '../imports/ui/Login.jsx';
-import ClassView  from '../imports/ui/ClassView.jsx';
-import App       from '../imports/ui/App.jsx';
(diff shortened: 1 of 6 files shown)


SamChou19815:
@julian-londono have you figured out the root cause of missing names?

julian-londono:
@SamChou19815 the payload coming back from Google just doesn't have them in it, still trying to figure out what makes it different

julian-londono:
I've made a fix to be able to continue to add users to the DB even if we fail to get their first and last names.  Will go this route indefinitely.

PR #102, opened 2019-10-19
Results Page Prototype

jjc387:
### Summary 

This pull request is the first step towards implementing the results page feature.

- [x] created a page to list the search results
- [x] created the component to hold each search result
- [x] modified the react-summary-gauge-2 library to help implement the component holding each search result such that it looks like the proposed design

### Test Plan 

Currently, nothing routes to my page, so I set a state variable such that whenever the meteor server starts, my page is mounted. I pass in a hard-coded result instance list to test if it maps each class instance to a component and displays all the classes. I look at the page and see if it resembles the proposed design.

### Notes 

Not sure if the react-summary-gauge-2 library was correctly updated.

### DIFF
--- a/.github/workflows/ci-workflow.yml
+++ b/.github/workflows/ci-workflow.yml
@@ -1,5 +1,5 @@
 name: CI
-on: pull_request
+on: push
 
 jobs:
   build:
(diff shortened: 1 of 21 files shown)


SamChou19815:
@julian-londono this diff is getting big and should be merged soon.

PR #117, opened 2019-11-14
Remove unused sendFeedback function

SamChou19815:
### Summary

This function is never used. It also looks dangerous since it's using jQuery.

### Test Plan

Proof: https://github.com/cornell-dti/course-reviews-react-2.0/search?q=sendFeedback&unscoped_q=sendFeedback


SamChou19815:
@julian-londono

PR #119, opened 2019-11-24
Stats - reviews over time chart 

AkashAryal:
Summary:
This pull request adds a chart component that plots the reviews over time of the top 15 majors to the admin dashboard. This allows for an easier way to determine the growth of our site by more easily comparing number of reviews.

Mini changes:
Updated accordion to start closed.

Test Plan:
Tested locally. Can view the chart in the admin dashboard and determine correctness.

Notes:
Chart was created using chartkick (https://chartkick.com/). 
Installed through npm.

Nothing in this pull request should cause problems to existing parts of the site.

### DIFF
--- a/imports/ui/Accordian.jsx
+++ b/imports/ui/Accordian.jsx
@@ -9,44 +9,44 @@ export default class Accordian extends Component{
   getRandNum(){
     return Math.floor((Math.random() * 1000000) + 0);
   }
+
   render(){
     let accId = "accordion"+this.getRandNum();
     let headingOneId = "headingOne"+this.getRandNum();
     let collapseOneId = "collapseOne"+this.getRandNum();
 
-      return(
-        <div className="panel-group" id={accId} role="tablist" aria-multiselectable="true">
-          <div className="panel panel-default">
-            <div className="panel-heading" role="tab" id={headingOneId}>
-              <h4 className="panel-title">
-                <a data-toggle="collapse" data-parent={"#"+accId} href={"#"+collapseOneId} aria-expanded="true" aria-controls={collapseOneId}>
-                  {this.props.title}
-                </a>
-              </h4>
-            </div>
-            <div id={collapseOneId} className="panel-collapse collapse" role="tabpanel" aria-labelledby={headingOneId}>
-              <div className="panel-body">
-              <table className="table table-striped">
-                <thead>
-                  <tr>
-                    <th scope="col">{this.props.col1}</th>
-                    <th scope="col">{this.props.col2}</th>
-                  </tr>
-                </thead>
-                <tbody>
-                  {
-                    this.props.data.map(function(data) {
-                      return(
-                        <tr key={data._id1}>
-                          <th scope="row" key={data._id2}>{data._id}</th>
-                          <td key={data._id3}>{data.total}</td>
-                        </tr>
-                      )
-                    })
-                  }
-                </tbody>
-              </table>
-              </div>
+    return(
+      <div className="panel-group" id={accId} role="tablist" aria-multiselectable="true">
+        <div className="panel panel-default">
+          <div className="panel-heading" role="tab" id={headingOneId}>
+            <h4 className="panel-title">
+              <a data-toggle="collapse" data-parent={"#"+accId} href={"#"+collapseOneId} aria-expanded="true" aria-controls={collapseOneId}>
+                {this.props.title}
+              </a>
+            </h4>
+          </div>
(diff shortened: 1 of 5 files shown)


imports/ui/Accordian.jsx
```diff
@@ -9,50 +9,50 @@ export default class Accordian extends Component{
...
     return Math.floor((Math.random() * 1000000) + 0);
   }
-  render(){
-    let accId = "accordion"+this.getRandNum();
-    let headingOneId = "headingOne"+this.getRandNum();
-    let collapseOneId = "collapseOne"+this.getRandNum();
+    render(){
+      let accId = "accordion"+this.getRandNum();
+      let headingOneId = "headingOne"+this.getRandNum();
+      let collapseOneId = "collapseOne"+this.getRandNum();
```
SamChou19815: The indentation looks messed up. Your indentation should be consistent with the rest of the code to avoid large diff and merge conflicts. Similar problem below.

imports/ui/Statistics.jsx
```diff
@@ -12,17 +20,46 @@ export default class Statistics extends Component{
...
+      chartData: []
     }
     this.howManyEachClass();
     this.howManyReviewsEachClass();
     this.totalReviews();
+    this.getChartData();
+  }
+
+  getChartData(){
+    data=[];
```
SamChou19815: Declaring a variable without `var`, `let` or `const` will create a *global variable*!
You should never create global variable.
You should use `let` or `const`.

imports/ui/Statistics.jsx
```diff
@@ -1,6 +1,14 @@
 import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { withTracker } from 'meteor/react-meteor-data';
+import { Reviews } from '../api/dbDefs.js';
+import { Classes } from '../api/dbDefs.js';
+import { Students } from '../api/dbDefs.js';
 import Accordian from './Accordian.jsx';
 
+import { LineChart, PieChart } from 'react-chartkick';
+import 'chart.js';
```
SamChou19815: Where is the place you are using this import?

server/methods.js
```diff
@@ -498,18 +499,112 @@ Meteor.methods({
...
+          'classSub': 1,
+          '_id': 0,
+          'classNum': 0
+        }).fetch()[0]; //finds the class corresponding to "KyeJxLouwDvgY8iEu" ex: cs 2112
+        let timeStringYMD = new Date(new Date().setDate(new Date().getDate() - i)).toISOString().split("T")[0];
+        if (retArr.includes(sub.classSub)) { //if thos review is one of the top 15 we want.
+          //date of this review minus the hrs mins sec
+          if (hashMap[sub.classSub] == null) //if not in hm then add
+            hashMap[sub.classSub] = {
+              [timeStringYMD]: data.total
```
SamChou19815: This extra computed property code looks extra.
```js
{ timeStringYMD: data.total }
```
should be enough.

imports/ui/Statistics.jsx
AkashAryal: Its used to create the line chart in line 97.

server/methods.js
AkashAryal: I need the value of timeStringYMD to be the field rather than the string "timeStringYMD." At the end hashMap is supposed to look something like this: 

{"cs": {date1: totalNumAtDate1, date2: totalNumAtDate2, ...}, ... }

imports/ui/Statistics.jsx
SamChou19815: Isn't it already handled by
```js
import { LineChart, PieChart } from 'react-chartkick';
```

AkashAryal: I thought so as well but the chart kick website (https://chartkick.com/react) in their quick start section had both listed. If I remove the chart.js import then chart does not load.

imports/ui/Accordian.jsx
```diff
@@ -9,6 +9,7 @@ export default class Accordian extends Component{
   getRandNum(){
     return Math.floor((Math.random() * 1000000) + 0);
   }
+
   render(){
     let accId = "accordion"+this.getRandNum();
```
ashneeldas2: Using a random number generator to come up with an ID can lead to insecurities, especially with an increasing number of ID's; You can use an alternate solution to ensure that you don't result in collisions.
@julian-londono

julian-londono:
Rebased onto master to avoid conflicts, force-pushed afterwards

PR #133, opened 2020-02-11
Made admin chart's range and step changeable

AkashAryal:
### Summary 
This pull request makes the admin chart's range and step changeable. Currently the chart shows data of reviews over a 1 yr interval with a step of 14 days. Now these numbers can freely be manipulated.  

### Test Plan 
Tested locally by changing the step and range numbers and verifying the result.


julian-londono:
Awesome job @AkashAryal

PR #144, opened 2020-03-17
Slack integration

khemritolya:
### Summary

This pull request (finally) adds a slack integration to CU Reviews. This has been updated from something we didn't merge a while ago, and which was 200 odd commits and a major rewrite behind master. Getting typescript to work with this is an interesting experience. I ran into what I think was [this](https://github.com/slackapi/node-slack-sdk/issues/948) issue on the slack api node sdk.

Features
- Get notified about new reviews users are posting!
- Approve reviews from within slack!

### Test Plan

TODO: integrate with test suite

### Notes

TODO: Typescript has issues with trying to work. We may never know why. Re-building the lockfile like [this](https://stackoverflow.com/questions/63777580/typing-express-nestjs-start-project-errors) suggests causes new and exotic errors in code which was correct before.

<details>
             <summary> Here's the old PR text: </summary>

### Summary

This pull request (finally) adds a slack integration to CU Reviews.

Features
- Get notified about new reviews users are posting!
- Approve reviews from within slack!

### Test Plan

Tested locally: Created a new slack environment for testing, added reviews locally, approved reviews within other environment, reviews appeared on local CU Reviews instance. Additionally tested both file not found and invalid path for env file, which made the server behave like before, allowing for local development without running a bot.

### Notes
The slack API is very strange. See the large block comment infront of the initializeBot function if you care about messing with it.

There are no breaking changes. If the bot lacks the required environment variables, it notifies you about this fact, but nothing happens.
</details>


khemritolya:
A lot has changed.

PR #149, opened 2020-04-12
Database update

khemritolya:
### Summary

This is a PR to manage the issue of reviews being submitted during the Covid-19 situation perhaps being unrepresentative of the true nature of the class.

As a result, this PR adds some functionality that addresses this issue:
1. Marks new reviews as being 'virtual' (which could be extended to online classes in the future as well)
2. When calculating average scores, allows us to provide a weight on 'virtual' scores from 0 (ignore) to 1 (count the same as a normal review). After some deliberation the team chose weight = 0. This is easily changeable in the future if we change out minds.

### Test Plan 

Tested locally. Chose a new class, added virtual reviews. Confirmed it showed the average of the virtual reviews. Added non-virtual reviews and more virtual reviews. Confirmed it always only showed the average of non-virtual reviews, as per above.

### Notes
A note on behavior: When a class has no non-virtual reviews, the gauges use the averaged virtual score regardless of the relative weighting of virtual to non-virtual reviews. This may be subject to change.
 
This PR also includes a minor bugfix. There was an issue on my machine, where search indices where not defined on the classes and subjects collections. I added some code to ensure that an index is always defined.

### DIFF
--- a/imports/api/dbDefs.js
+++ b/imports/api/dbDefs.js
@@ -101,6 +101,8 @@ Reviews.schema = new SimpleSchema({
     // The full functional code for counting reviews can be found on the following branch:
     // review-counting-feature
     // memberReferral: { type: String, optional: true }, // DTI member referral for review contest
+    virtual: { type: Boolean }, // whether or not this review is for a class was taught virtually at least part of the time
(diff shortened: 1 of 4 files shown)


khemritolya:
Closed in favor of https://github.com/cornell-dti/course-reviews-react-2.0/pull/167

PR #155, opened 2020-04-13
removing placeholder text from Navbar/Searchbar in results and review page

AkashAryal:
### Summary 
This pull request removes placeholder text in the searchbar's input field for the results and review page

### Test Plan
Checked locally by navigating to the respective pages


julian-londono:
Thanks for the fix @AkashAryal !

PR #157, opened 2020-04-15
Blue redesign

julian-londono:
### Summary 

This PR implements the complete redesign of the results page and class review page.  Changes include:
- Font changes
- Color scheme changes
- Functionality changes (ability to sort in class review page)
- Misc style changes
- CSS class name refactoring

Real results page VS designed results page:

Screen Shot 2020-04-14 at 8 20 45 PMScreen Shot 2020-04-14 at 8 21 06 PM

Real class review page vs designed class review page:

Screen Shot 2020-04-14 at 8 21 21 PMScreen Shot 2020-04-14 at 8 21 33 PM

### Test Plan 

Will test extensively on dev site before merging and pushing to prod.

### Notes 

We will create documentation/guidelines on CSS best practices going forward to keep the site's styling maintainable going forward. 

### Breaking Changes  

- [x] Known issue: Reporting a review currently does not make it disappear from the class page but _does_ mark it as reported

- [x] Bug in professors listed in preview card on results page.

- [x] DTI logo missing in the authentication pop-up

- [x] Double scrollbars in the class review page


SamChou19815:
In this future, we should avoid creating feature branches like this. Everything should be targeting `master` and we can avoid a big merge like this.

rockon999:
> In the future, we should avoid creating feature branches like this. Everything should be targeting `master` and we can avoid a big merge like this.

Creating two branches, one for production and one for development could ease these issues if the team wants to track a redesign in git while maintaining a production branch.

julian-londono:
> > In the future, we should avoid creating feature branches like this. Everything should be targeting `master` and we can avoid a big merge like this.
> 
> Creating two branches, one for production and one for development could ease these issues if the team wants to track a redesign in git while maintaining a production branch.

We used to have a `dev` branch but were advised to move away from this workflow set up.  So in this case `blue-redesign` became our `dev` branch for this massive overhaul.  Not sure what the ideal workflow would have been for this, thoughts @rockon999 @SamChou19815 ?

SamChou19815:
@julian-londono every pull request should be a transition like:

```
deployable master -> [your PR] -> a new deployable master with PR's changes.
```

In this case of the design overhaul, it's not possible that each PR that implements some redesign is deployable. Then you should set a flag that toggle the display new design. Therefore, we can still land emergency fixes like #152 to prod by turning the flag of new design off.

rockon999:
> > > In the future, we should avoid creating feature branches like this. Everything should be targeting `master` and we can avoid a big merge like this.
> > 
> > 
> > Creating two branches, one for production and one for development could ease these issues if the team wants to track a redesign in git while maintaining a production branch.
> 
> We used to have a `dev` branch but were advised to move away from this workflow set up. So in this case `blue-redesign` became our `dev` branch for this massive overhaul. Not sure what the ideal workflow would have been for this, thoughts @rockon999 @SamChou19815 ?

Yes, there are pitfalls to a permanent development branch. Mainly, like in this case, you end up with a massive merge-in.

In my opinion, the cleanest way to do this is to version your branches...

* `master` is the development branch
* When you are ready for a new release (large redesigns, new features, etc) you create a new branch off of master called `v2`, `v3` or whatever version marker you want.
* If you need to port code back into the older branches or apply a fix you can.
* Use tags to indicate "subversion" (v2.1)

Alternatively, you can keep the current release in a `production` branch and only version older branches: `v1` -> `v2` -> `production` (v3) -> `master` (v4)

julian-londono:
This calls for a toast @draycfarley @AkashAryal 🎉

PR #167, opened 2020-04-21
Database Update

khemritolya:
### Summary

This is a PR that branches from the correct branch. The previous version was here: https://github.com/cornell-dti/course-reviews-react-2.0/pull/149 This PR has all these features and more.

This is a PR to manage some issues with the database.

1. There is the issue that reviews being submitted during the Covid-19 situation perhaps being unrepresentative of the true nature of the class.
2. We want to be able to search and group by professors

As a result, this PR adds some functionality that addresses these issues:
1. Marks new reviews as being 'virtual' (which could be extended to online classes in the future as well)
2. When calculating average scores, allows us to provide a weight on 'virtual' scores from 0 (ignore) to 1 (count the same as a normal review). After some deliberation the team chose weight = 0. This is easily changeable in the future if we change out minds.
3. Adds a "professors" collection to mongo to allow us to search by professor in the future
4. Adds a script to generate the professors collection from the existing database

### Test Plan 

Tested locally. 

For the Virtual Class Marking: Chose a new class, added virtual reviews. Confirmed it showed the average of the virtual reviews. Added non-virtual reviews and more virtual reviews. Confirmed it always only showed the average of non-virtual reviews, as per above.

For the Professors-building Script: Ran script, checked script output which shows that script linked the correct professors and classes using the "_id" field.

### Notes
A note on behavior: When a class has no non-virtual reviews, the gauges use the averaged virtual score regardless of the relative weighting of virtual to non-virtual reviews. This may be subject to change.
 
How to run/test the Professors building script:
 1. Have a local mongodb open: `mongod --dbpath ./mongo --port 3001`
 2. Open the mongodb shell for this local instance `mongo --port 3001`
 3. run `load(".scripts/buildprofs.js")`

About Bugs found while working on this PR: 
- There was an issue on my machine, where search indices where not defined on the classes and subjects collections. I added some code to ensure that an index is always defined.
~~There is an issue where an "entry" field is not defined in a backend function for the graphs. I was unable to diagnose this issue, but added some code to avoid errors. Someone should probably fix this, as it does not occur on master.~~ This has been fixed on master. Reverted.


julian-londono:
Last thing to add here is the ability to search by professor tags in the searchbar - will implement tomorrow

julian-londono:
Your work was not done in vain @khemritolya !

PR #168, opened 2020-04-21
Remove main X in professor dropdown

julian-londono:
### Summary 

This PR just removes the main "X" from the professor dropdown in the form while leaving the individual X's so that users can still remove selected professors

### Test Plan 

Tested locally, it gone.


julian-londono:
Taken care of in #169

PR #169, opened 2020-04-23
Smaller screen reactive

julian-londono:
### Summary 

This PR makes the results and class review pages more reactive as well as fixing slight spacing issues.

### Test Plan 

Tested screen sizes locally.

### Notes 

Note: This is only a quick fix while the true mobile implementation comes


imports/ui/Gauge.jsx
```diff
@@ -55,15 +55,16 @@ export default class Gauge extends Component{
...
-            <h1 className={this.gauge_text_top}>
-                {this.state.rating}
-            </h1>
+          <div className={"col-md-6 col-sm-6 col-xs-6 no-side-padding gauge-text-container " + this.gauge_text_padding}>
+            <p className="gauge-text">
+                <span className={this.gauge_text_top}>{this.state.rating}</span>
+                <span className="gauge-text-bottom ">{this.props.text}</span>
+            </p>
             <h1 className="gauge-text-bottom">
-              {this.props.text}
```
SamChou19815, line 63: Why this change?

julian-londono: - `h1` --> `p` was just a preference

- putting both into the same `p` tag surrounded by their own `span` tags was for vertical centering using auto margins to work

SamChou19815: Then why do we left an empty h1 here?

julian-londono: Ahh totally missed this `h1` tag, will remove now, thanks for the catch!

PR #172, opened 2020-04-25
Insert previous query to searchbar

khemritolya:
### Summary

This feature adds the previously searched query to the search bar on a page

### Test Plan

Went to start page. Placeholder text was still present. No default text seen. Searched for something. Verified that the correct query showed up.


khemritolya:
Sorry about the issues. Everything should be fixed.

julian-londono:
Awesome functionality, clean approach @khemritolya 👍

PR #173, opened 2020-04-26
[WIP] Better scraping

khemritolya:
### Summary

This pull request is the beginnings of the work to improve web scraping. I made it mostly because I was curious about the type-script work being done on the meteor-shower branch.

I came up with an alternative implementation for the web scraper. In particular, I attempted to improve runtime with the use of bulk database writing.

### Test Plan

Ran the code with a little tweak each time (see line 16 in scrapers.ts). For sample size n=1 (to be improved):

addAllCoursesNew => Test time: 64 s
addAllCoursesOld => Test time: 55 s

Currently, it seems that the old approach is **faster** than the batched writing. This will be investigated further.

### Notes / Breaking Changes

This work revealed an underlying issue in the database migration work of the meteor shower branch. Mongoose is not expecting an explicit _id field in the collections definition, which was causing db writes to fail on my machine. I removed the explicit _id field, and confirmed that it was in fact present implcitly in the collection. See the changes to dbDefs.ts. **NOTE** this *might* cause problems all over the backend, which is expecting _id's to be strings rather than ObjectIds. Whether or not this actually has an effect is tbd.

There are several issues with the analysis thus far. The sample size is obvious. Additionally, the vast majority of the time seems to be spent on fetching data from Cornell's API, which impacts the ability for us to make any gains regarding efficieny by improving database write speed.

### DIFF
--- a/server/scrapers.ts
+++ b/server/scrapers.ts
@@ -1,0 +1,216 @@
+import axios from "axios";
+import { Classes, Subjects, Class } from "./dbDefs";
+
+// Due to the asynchronous nature of the program, any times measures are unreliable
+
+function time() {
+    return Date.now() / 1000;
+}
+
+export async function runTests() {
+    console.log("Running tests");
+    // For the purposes of testing, drop our collections
+    await Subjects.deleteMany({}).exec();
+    await Classes.deleteMany({}).exec();
+
+    const runTime = await addAllCoursesOld(["SP19", "FA19"]);
+    console.log("Test time: " + runTime);
+}
+
+/*
+ * A new scraping function. Uses batch writing to maybe improve runtime.
+ */
+async function addAllCoursesNew(semesters: string[]) {
+    const begin_time = time();
+
+    console.log("Query over semesters: " + semesters)
+
+    // Await the joint completion of fetching all information for all subjects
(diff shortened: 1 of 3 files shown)


SamChou19815:
@khemritolya Your branch is based upon the `meteor-shower` branch a while ago. Try rebase on top of the most recent one. I might already fixed some of the problems.

khemritolya:
Sorry about the git problems

server/scrapers.ts
```diff
@@ -0,0 +1,216 @@
...
+
+    let classRes = await Classes.collection.insertMany(classDocs);
+
+    if (classRes.result.ok != 1) {
+        throw new Error("Unable to write classes to db!");
+    } else {
+        console.log("Done writing " + classRes.result.n + " classes to db!");
+    }
+
+    console.log(await Classes.find({}).exec());
```
SamChou19815, line 134: > Currently, it seems that the old approach is faster than the batched writing. This will be investigated further.

@khemritolya Is it possible that it's caused by this line in the new approach?

khemritolya: It very well could be. I'll look into it when I get around to doing a proper analysis over the weekend.

PR #182, opened 2020-05-14
Express fixes

SamChou19815:
### Summary 

- Fixed some linter errors by @SamChou19815, @khemritolya, @draycfarley
- Migrated several components to typescript by @AkashAryal (future task: replace `any` with better types)
  - `CourseCard`
  - `CourseReviews`
  - `FilteredResult`
  - `Gauge`
  - `Statistics`
  - `UpdateReview`
- Added some missing `await` in backend code by @khemritolya 

### Test Plan 

- [x] Build passes
- [x] Existing tests pass
- [x] Fewer linter errors
- [ ] Test in staging


server/methods.ts
```diff
@@ -410,7 +426,7 @@ Meteor.methods({
       if (ticket && ticket.email) {
         const user = await Meteor.call<Student | null>('getUserByNetId', ticket.email.replace('@cornell.edu', ''));
         if (user) {
-          return user.privilege === 'admin';
+          return user.privilege !== 'admin';
```
SamChou19815, line 429: This change looks wrong.

khemritolya: It is. Made the change locally because testing admin functions is a pain. I can't belive I was stupid enough to push it.

khemritolya:
For the test plan, it might be nice to have unit tests for all these methods. I'm not sure how we would do that, given the fact that many of these methods require a google token. It might be nice to override them with an environment variable, but I'm not sure how we would do that securely.

SamChou19815:
> For the test plan, it might be nice to have unit tests for all (or most) of the backend methods. I'm not sure how we would do that, given the fact that many of these methods require a google token. It might be nice to override them with an environment variable, but I'm not sure how we would do that securely. It would also entail a lot more changes on the backend.

Doing that won't be hard:

```
// methods.test.ts
process.env.MOCK_GOOGLE_TOKEN = 'true';

// method.ts
async login(token: string) {
  if (process.env.MOCK_GOOGLE_TOKEN === 'true') {
    if (token === 'HARDCODED_TOKEN_FOR_ADMIN') {
      return MOCK_ADMIN_USER;
    }
    else if (token === 'HARDCODED_TOKEN_FOR_TEST_USER') {
      return MOCK_NORMAL_USER;
    } else {
      return reportTokenError();
    }
  }
  // normal login flow
}
```

However, it's better to do it in a separate PR to avoid inject too many things into this already big PR.

PR #186, opened 2020-09-09
Resolving some graph errors

AkashAryal:
### Summary 

This pull request resolves the issues that caused the graph to not load.
The issue was related to promises in the backend and changes to mongo that did not allow certain forms of queries.

This pull request also resolves the `b is not a function` when navigating to the admin page by updating `react-google-login`

### Test Plan 
Checked locally by navigating to /admin

### Notes 
There was a lot of issues with the following proxy error:
```
Proxy error: Could not proxy request /methods/getReviewsOverTimeTop15 from localhost:3000 to http://localhost:8080.
See https://nodejs.org/api/errors.html#errors_common_system_errors for more information (ECONNRESET).
```
As of this pull request, this error appears only once the first time you navigate to /admin. Previously, the moment you went to /admin the chart would begin to load with default values. Now you will have to press the load chart to get the chart to load. 

Not exactly sure as to the cause of this. It could be a timing thing with the connection being reset/cutoff due to the chart taking several seconds to load.

Edit: Moving the call to the function that gets chart data (getChartData) from the constructor to componentDidMount seems to have "solved" the problem (in the sense that I don't get the error message anymore). Not sure if that's just a work around or the actual root

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ mongo_backups/*
 db
 .meteor
 mongo/
+*error.log
--- a/client/package.json
+++ b/client/package.json
@@ -15,7 +15,7 @@
     "react-chartkick": "^0.4.0",
     "react-circular-progressbar": "^2.0.3",
     "react-dom": "^16.13.1",
-    "react-google-login": "^5.1.2",
+    "react-google-login": "^5.1.21",
     "react-loading-animation": "^1.4.0",
     "react-router-dom": "^5.1.2",
     "react-scripts": "3.4.1",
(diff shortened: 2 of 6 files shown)


julian-londono:
So @AkashAryal was correct about a timeout causing `Could not proxy request /methods/getReviewsOverTimeTop15`.  I've updated the timeout value to fix that.  

We still get the proxy error in the following situations:
- Whenever any call to the backend throws and therefore a promise is rejected
- On the admin page when one tries to or when one tries to approve or reject a review before the chart loads initially

This means we need better error handling on the back-end with and maybe a refactor of admin page loading logic - however I think this PR is ready to merge and the changes mentioned should come in different PRs

julian-londono (approved):

PR #188, opened 2020-09-10
Remove jquery from deprecated code

julian-londono:
### Summary 

This PR removes jQuery as a dependency in `/client/index.ts` and therefore completely from the codebase

It was being used for a deprecated feature in a function that was no longer relevant - I've removed that function as well.

### Test Plan 

Tested typical workflow locally


AkashAryal (approved):
Looks good. Definitely don't need jQuery anymore.

PR #191, opened 2020-09-13
Admin Page Auth Issues

julian-londono:
### Summary 

This PR fixes a flaky step in the authentication of the `/admin` path.

Issue was: the Session variable `adminlogin` was not being reset if a `token` was outdated - this resulting in users being able to access the admin page but leading to issues once on the page because their admin token was out of date

### Test Plan 

Tested workflow locally. Will test on dev server.

### Notes 

I advise that after this is merged, all current branches should be rebased on top of `master` to make sure these issues are fixed while working on updates to the admin page


draycfarley (approved):
Looks good to me

AkashAryal:
Ayy. I dont have to clear local storage every few minutes to access /admin.

PR #192, opened 2020-09-14
Error handling

julian-londono:
### Summary 

This PR adds error handling on the back-end to all functions in `server/methods.ts`.  These exist in the form of `try/catch` statements that prints <ins>on the server's log</ins> where a function fails and the full `error` object caught as a result.

Ideally, this error handling should be automated for all calls to the back - but since we're still using the `server/shim.ts` and `client/meteor-shim.ts` files for these calls, I figured we could implement this at the individual method level for now.

### Test Plan 

Tested locally - found a successful case where a function failed and was caught and logged on the backend.  Will test functionality on dev site

### DIFF
--- a/server/methods.ts
+++ b/server/methods.ts
@@ -82,51 +82,59 @@ Meteor.methods({
    * Returns 1 on a success
    */
   async insert(token, review, classId) {
-    if (token == undefined) {
-      console.log("Error: Token was undefined in insert");
-      return 0;
-    }
-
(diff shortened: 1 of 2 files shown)


AkashAryal (approved):
This will definitely make debugging with the backend much easier.

julian-londono:
@AkashAryal I've updated the description of the PR to address what you initially mentioned about error handling with regards to the `shim`

PR #193, opened 2020-09-14
Metrics update after undoing report and reviews disappear after reporting

draycfarley:
### Summary 

This pull request updates the back end so that "updateCourseMetrics" is called for when reviews are removed from the admin page so the metrics are updated. In addition, when a user reports a review, it will disappear from the UI without refreshing

### Test Plan 

Tested locally through front end and print statements.

### DIFF
--- a/client/src/ui/CourseReviews.tsx
+++ b/client/src/ui/CourseReviews.tsx
@@ -89,6 +89,13 @@ export class CourseReviews extends Component<Props, State> {
   reportReview = (review: ReviewType) => {
     Meteor.call('reportReview', review, (error: any, result: 1 | 0) => {
       if (!error && result === 1) {
+        let idx = 0;
+        console.log(review._id);
+        while (idx < this.state.reviews.length && this.state.reviews[idx].key !== review._id) {
+          idx++;
+        }
+        console.log(idx);
+        this.setState({reviews: this.state.reviews.slice(0, idx).concat(this.state.reviews.slice(idx+1)) });
         console.log("reported review #" + review._id);
       } else {
         console.log(error)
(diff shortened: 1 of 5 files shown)


client/src/ui/CourseReviews.tsx
```diff
@@ -89,6 +89,11 @@ export class CourseReviews extends Component<Props, State> {
   reportReview = (review: ReviewType) => {
     Meteor.call('reportReview', review, (error: any, result: 1 | 0) => {
       if (!error && result === 1) {
+        let idx = 0;
+        while (idx < this.state.reviews.length && this.state.reviews[idx].key !== review._id) {
+          idx++;
+        }
+        this.setState({reviews: this.state.reviews.slice(0, idx).concat(this.state.reviews.slice(idx+1)) });
```
julian-londono, line 96: I hate that this is one of the simplest ways to remove an element by index in JS but don't think there's another simpler way 👍

server/methods.ts
```diff
@@ -248,14 +248,13 @@ Meteor.methods({
...
         await Classes.updateOne({ _id: courseId },
           {
             $set: {
               // If no data is available, getGaugeValues returns "-" for metric
-              classDifficulty: (state.diff !== "-" ? Number(state.diff) : null),
-              classRating: (state.rating !== "-" ? Number(state.rating) : null),
-              classWorkload: (state.workload !== "-" ? Number(state.workload) : null),
+              classDifficulty: (state.diff !== "-" && !isNaN(state.diff) ? Number(state.diff) : null),
+              classRating: (state.rating !== "-" && !isNaN(state.rating) ? Number(state.rating) : null),
+              classWorkload: (state.workload !== "-" && !isNaN(state.workload) ? Number(state.workload) : null),
```
julian-londono, line 257: Good call, this was a frequent issue

julian-londono (dismissed):

julian-londono:
Wait didn't mean to approve yet, testing locally now

julian-londono (changes requested):
This correctly addresses a review disappear from the class page when reported, could this be amended to address a review disappearing from the admin page when approved/rejected? @draycfarley

julian-londono (approved):
As discussed, the similar functionality will go in a separate PR.  This PR checks out on my local.  Can you merge when you get the chance @draycfarley

PR #194, opened 2020-09-15
Profanity Filter for client and server

AkashAryal:
### Summary 
This pull request adds a profanity filter for the client and server.

### Test Plan 
Tested locally with sample input. The main function used to implement the profanity filter is the "bad-words" package's isProfane function. Thus, I didn't add any unit tests.


khemritolya (dismissed):
The core idea is good. From looking over it myself I have 3 concerns however. 

First is this: 

Screenshot from 2020-09-15 15-32-19. 

I think that it is picking up on the profanity in "class". This is probably very much unintented behavior. There are probably other words like this. Perhaps we should have an allow list for certain words which overrides the profanity filter somehow. There might be better solutions, however.

Second is that it seems that we are just copy-pasting these profanity bits from another git repo. We should probably attribute the source somehow, or pull them remotely, and if that fails fall back on the static list. One possible advantage of pulling remotely is that we could change the source easily in the future, and receive any automatic updates. Food for thought I guess.

Thrid, we might also want to integrate this with jest, and write some tests for it, so that something like issue 1 doesn't happen again.

AkashAryal:
Yea that's certainly not intended lol but I'm not sure how to best proceed. One method that comes to mind is that I could assert that there must be a space or special character after the bad word. This way "class" would be allowed since there is a "l" before it. But this has its own issues too since the user can just concat a random character before/after the bad word and it would pass the filter. Also this would mean removing spaces wouldn't work which means "b a d w o r d" will pass. Other methods that come to mind also fail in certain cases. Not sure if there is a way to do this without a whitelist. 

I like the idea of a whitelist but creating one that includes everything will be quite difficult (ex: i just learned that shitake mushrooms were a thing lol). I tried looking online but couldn't find any common false positive words either. 

For the list, I'm using the one linked in asana (txt file). I can definitely add the link as a comment. I'm not sure how much the list will change in the future so I don't think its worth writing a parser that downloads a source txt from the web and then parses it into an array. I guess we could though if we end up having to change our black list alot.

I guess for now its better to allow some false negatives than prevent reviews from being submitted. So i can implement the above method of asserting spaces for now? What do you think @julian-londono?

julian-londono:
I think there's good initial intuition with removing space - however, profanity filtering has had a ton of research done and we definitely don't wanna re-invent the wheel.  With that said, you might wanna look into existing libraries for this stuff:
- Most popular: [bad-words](https://github.com/web-mech/badwords#readme)
- An alternative: [swearjar-node](https://github.com/raymondjavaxx/swearjar-node)

One of these seems like a good starting point for us - I apologize for not recommending something like this before!

NOTE: remember we're now using yarn so installing the first one for example would be:
```bash 
# In `/common`, because this would be used by both front-end and backend
yarn add bad-words
```

client/src/ui/Form.jsx
```diff
@@ -368,121 +374,122 @@ export default class Form extends Component {
...
+          <form className="form" onSubmit={this.handleSubmit.bind(this)} ref={this.formElement}>
+            <p className="form-header-text">Leave a Review</p>
+            {this.props.searchBar && <SearchBar formPopupHandler={this.setCourseIdInSearchBar} isPopup={true} />}
+            <div className="row form-textbox-row">
+              <textarea ref={this.textArea} className={"form-input-text" + (err.text || err.textEmpty ? "error" : "")} type="text" value={this.state.text}
+                onChange={(event) => this.handleTextChange(event)}
+              />
+              <div ref={this.emptyMsg} className={err.textEmpty ? "form-field-error" : "hidden"}>Please add text to your review!</div>
+              <div className={err.text && this.state.text !== "" ? "form-field-error" : "hidden"} id="errorMsg" >Your review contains illegal characters, please remove them.</div>
+              <div className={!err.text && !err.textEmpty && err.profanity ? "form-field-error" : "hidden"} id="errorMsg" >Your review contains profanity, please edit your response.</div>
```
draycfarley, line 388: Could be refactored in the future to a map function on a list of errors, whether they're present, and the error message instead of hard coding a div for each one.

draycfarley (approved):

common/Profanity.js
```diff
@@ -0,0 +1,10 @@
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const Filter = require('bad-words');
+
+const filter = new Filter();
+
+const includesProfanity = (s) => filter.isProfane(s);
+
+module.exports = {
+  includesProfanity,
+};
```
julian-londono, line 10: Lowercase this file name to keep consistency in this directory.  Uppercase file names make it look like it should be a library.

server/methods.ts
```diff
@@ -2,6 +2,7 @@ import { getGaugeValues, getCrossListOR } from 'common/CourseCard';
 import { OAuth2Client } from 'google-auth-library';
 import { TokenPayload } from 'google-auth-library/build/src/auth/loginticket';
 import shortid from 'shortid';
+import { includesProfanity } from "common/Profanity";
```
julian-londono, line 5: Changing case of `Profanity` will apply here too

server/methods.ts
```diff
@@ -84,20 +85,25 @@ Meteor.methods({
...
 
-    if (!ticket) return 0;
+    if (!ticket) return { resCode: 0, errMsg: "Missing verification ticket" };
 
     if (ticket.hd === "cornell.edu") {
       // insert the user into the collection if not already present
       await Meteor.call("insertUser", ticket);
 
+      if (review.text !== null && includesProfanity(review.text)) {
+        console.log("profanity detected in review.");
```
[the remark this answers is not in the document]
julian-londono: To keep consistency, go through and add `// eslint-disable-next-line no-console` to the lines above where you used a `console.log` - suppresses the linter warnings

julian-londono (changes requested):
Good idea returning an object with both `resCode` and `errMsg` to display.  A couple changes requested in comments.

julian-londono:
Had to rebase to account for changes on `master`. Before making any remaining changes, make sure you force pull locally: `git pull -f origin profanity-filter`

julian-londono (approved):

draycfarley:
test github slack comment

julian-londono:
Could you address the merge conflict in `client/src/ui/Form.jsx` @AkashAryal ?

client/src/ui/Form.jsx
```diff
@@ -251,14 +249,11 @@ export default class Form extends Component {
       !Session.get("courseId") ? this.props.course._id : Session.get("courseId"),
       (error, result) => {
         // if (!error && result === 1) {
-        if (error || result.resCode === 1) {
+        if (error || result === 1) {
```
julian-londono, line 252: Don't you need `result.resCode` (as opposed to `result` for your implementation of success/error messages on the user-side? @AkashAryal

AkashAryal: Yea that should be `result.resCode`.

PR #145, opened 2020-03-28
Meteor Shower: Washing Away Meteor

rockon999:
This pull request works to remove Meteor, in its entirety, from CU Reviews. It does so by supplying basic shims for Meteor's method and subscription APIs and migrating database calls to `mongoose`.

The client is moved to `create-react-app` in `client/` whilst the server uses express and Node.js in `server/`.

- [X] Convert client to `create-react-app`
- [X] Shim basic Meteor behavior
- [X] Use Yarn workspaces as a replacement mechanism of code sharing between frontend and backend

This is based on Sam's TypeScript work :heart: 

### Test Plan

To run the app:

1. check out the branch.
2. Install [yarn](https://classic.yarnpkg.com/lang/en/)
3. `yarn install` to install all dependencies
4. Put `MONGODB_URL=[SECRET_URL]` in `server/.env`
5. Start backend by `yarn workspace server start`
6. Start frontend by `yarn workspace client start`
7. Play with it on localhost:3000

Aggressively test for regressions against cureviews.org

### Notes

Once the `blue-redesign` goes into master, rebase and change the base to `master` on github.

Once this diff is done, we can close #118, close #123, close #127, close #135, since it contains all the fixes in those PRs.

PR #195, opened 2020-09-15
Improved Scraper 2

khemritolya:
### Summary 

Goal: After #145 , we don't know if our scraper works. This is an attempt to write a tested scraping implementation. The old one was also not easy to understand.

It's like #173, except our goal is to make to make a correcter scraping implementation rather than a faster scraping implementation. Fun Fact: Compared to the 15 mins it historically took, this is actually doing pretty well!

Frontend integration will be split out into a seperate PR.

### Test Plan 

All tests are in the file dbInit.test.ts. We create a mock endpoint, and then scrape from it. The scraper is split up into various utility functions, which are seperatly tested.

It was also tested on a clean db. You can also do so by running. This will automatically do everything for you. No need to fumble about with mongod or anything.
```bash
ALLOW_LOCAL=1 yarn workspace server start
```

### Notes

Added the ALLOW_LOCAL program option. Passing ALLOW_LOCAL=1 and no mongodb uri to connect to automatically now starts a blank in-memory mongo server, which is the populated using FA19 data. This process takes ~1-2 mins on my machine. This is useful to test features involving the db, if you are worried that this might cause issues with the staging db. In this case, it was used to manually verify that scraping worked. NOTE: there are still some errors, because the professors db is not properly initialized. TODO: Fix this.

### DIFF
--- a/server/dbInit.test.ts
+++ b/server/dbInit.test.ts
@@ -13,7 +13,10 @@ jasmine.DEFAULT_TIMEOUT_INTERVAL = 600000;
 let testServer: MongoMemoryServer;
 let serverCloseHandle;
 
-// Configure a mongo server and fake enpoints for the tests to use
+const testingPort = 27760;
+const testingEndpoint = `http://localhost:${testingPort}/`;
+
+// Configure a mongo server and fake endpoints for the tests to use
 beforeAll(async () => {
   // get mongoose all set up
   testServer = new MongoMemoryServer();
@@ -39,11 +42,9 @@ beforeAll(async () => {
     classDifficulty: 5,
   }).save().catch((err) => { console.log(err); });
 
-  console.log(await Subjects.findOne({ subShort: "GORK" }));
-
   // We need to pretend to have access to a cornell classes endpoint
   const app = express();
-  serverCloseHandle = app.listen(27760, async () => {});
+  serverCloseHandle = app.listen(testingPort, async () => {});
 
   app.get("/hello", (req, res) => {
     res.send("Hello world");
@@ -99,8 +100,6 @@ afterAll(async () => {
   serverCloseHandle.close();
 });
 
-const testingEndpoint = "http://localhost:27760/";
-
 describe('tests', () => {
   it("dbInit-db-works", async () => {
     expect((await Subjects.findOne({ subShort: "GORK" })).subShort).toBe("GORK");
@@ -108,7 +107,7 @@ describe('tests', () => {
   });
 
   // Does the internal testing endpoint exist?
-  it("dbInit-test-enpoint-exists", async () => {
+  it("dbInit-test-endpoint-exists", async () => {
     const response = await axios.get(`${testingEndpoint}hello`);
     expect(response.data).toBe("Hello world");
     expect(response.data).not.toBe("Something the enpoint is not to return!");
--- a/server/dbInit.test.ts
+++ b/server/dbInit.test.ts
@@ -25,16 +25,16 @@ beforeAll(async () => {
 
   await new Subjects({
     _id: "some id",
-    subShort: "GORK",
+    subShort: "gork",
     subFull: "Study of Angry Fungi",
   }).save();
 
   await new Classes({
     _id: "some other id",
-    classSub: "GORK",
+    classSub: "gork",
     classNum: "1110",
     classTitle: "Introduction to Angry Fungi",
-    classFull: "GORK 1110 Introduction to Angry Fungi",
+    classFull: "gork 1110 Introduction to Angry Fungi",
     classSems: ["FA19"],
     classProfessors: ["Prof. Thraka"],
     classRating: 5,
@@ -65,8 +65,8 @@ beforeAll(async () => {
       status: "success",
       data: {
         subjects: [
-          { descr: "Study of Fungi", descrformal: "The Study of Fungi", value: "GORK" },
-          { descr: "Study of Space", descrformal: "The Study of Where No One has Gone Before", value: "FEDN" },
+          { descr: "Study of Fungi", descrformal: "The Study of Fungi", value: "gork" },
+          { descr: "Study of Space", descrformal: "The Study of Where No One has Gone Before", value: "fedn" },
         ],
       },
     });
@@ -74,9 +74,9 @@ beforeAll(async () => {
 
   // Fake classes endpoint
   app.get("/search/classes.json", (req, res) => {
-    // simulate only having data for the GORK subject.
+    // simulate only having data for the gork subject.
     // see above
-    if (!req.originalUrl.includes("GORK")) {
+    if (!req.originalUrl.includes("gork")) {
       res.send({
         status: "failure",
       });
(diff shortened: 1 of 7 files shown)


server/dbInit.test.ts
```diff
@@ -0,0 +1,158 @@
...
+  });
+});
+
+afterAll(async () => {
+  await mongoose.disconnect();
+  await testServer.stop();
+  serverCloseHandle.close();
+});
+
+const testingEndpoint = "http://localhost:27760/";
```
julian-londono, line 102: Suggestion: make `27760` a `const` at the top since it's used multiple places

server/dbInit.test.ts
```diff
@@ -0,0 +1,158 @@
...
+  it("dbInit-db-works", async () => {
+    expect((await Subjects.findOne({ subShort: "GORK" })).subShort).toBe("GORK");
+    expect((await Classes.findOne({ classSub: "GORK", classNum: "1110" })).classSub).toBe("GORK");
+  });
+
+  // Does the internal testing endpoint exist?
+  it("dbInit-test-enpoint-exists", async () => {
+    const response = await axios.get(`${testingEndpoint}hello`);
+    expect(response.data).toBe("Hello world");
+    expect(response.data).not.toBe("Something the enpoint is not to return!");
```
julian-londono, line 114: Typo: "enpoint" --> "endpoint"

server/dbInit.test.ts
```diff
@@ -0,0 +1,158 @@
...
+import { fetchSubjects, fetchClassesForSubject, fetchAddCourses } from "./dbInit";
+
+// May require additional time for downloading 100 mb (!) worth of MongoDB binaries
+// **We might not want to run this with CI**
+jasmine.DEFAULT_TIMEOUT_INTERVAL = 600000;
+
+let testServer: MongoMemoryServer;
+let serverCloseHandle;
+
+// Configure a mongo server and fake enpoints for the tests to use
```
julian-londono, line 16: Typo: "enpoints" --> "endpoints"

server/dbInit.ts
```diff
@@ -1,6 +1,156 @@
...
+      }).save().catch((err) => null);
+
+      // db operation was not successful
+      if (!res) {
+        throw new Error();
+      }
+    }
+
+    return true;
+  })).catch((err) => null);
```
julian-londono, line 77: Would be helpful to print this error to the backend log (just `console.log` since this is already running on the server) instead of just tossing it

server/dbInit.ts
```diff
@@ -1,6 +1,156 @@
...
+          .catch((err) => null);
+
+        if (!res) {
+          console.log(`Unable to update class information for ${cl.subject} ${cl.catalogNbr}!`);
+          throw new Error();
+        }
+      }
+
+      return true;
+    })).catch((err) => null);
```
julian-londono, line 136: Same as above, print this error

julian-londono (changes requested):

khemritolya:
@julian-londono Should we make the frontend thing its own PR?

julian-londono:
> @julian-londono Should we make the frontend edits its own PR?

I'd say so yes

server/dbInit.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    classSub: "GORK",
+    classNum: "1110",
+    classTitle: "Introduction to Angry Fungi",
+    classFull: "GORK 1110 Introduction to Angry Fungi",
+    classSems: ["FA19"],
+    classProfessors: ["Prof. Thraka"],
+    classRating: 5,
+    classWorkload: 5,
+    classDifficulty: 5,
+  }).save().catch((err) => { console.log(err); });
```
julian-londono, line 43: Sorry should have clarified this before: for the sake of clarity and usefulness in the logs, could you please log an explanation of the log error maybe like this to all these logs?

https://github.com/cornell-dti/course-reviews-react-2.0/blob/f30c4ca2f22682e574e47f703c3fd1100d558f6b/server/methods.ts#L198

server/dbInit.ts
```diff
@@ -1,6 +1,165 @@
...
+    const subjectIfExists = await Subjects.findOne({ subShort: subject.value.toUpperCase() }).exec();
+
+    if (!subjectIfExists) {
+      console.log(`Adding new subject: ${subject.value}`);
+      const res = await new Subjects({
+        _id: shortid.generate(),
+        subShort: subject.value.toUpperCase(),
+        subFull: subject.descrformal,
+      }).save().catch((err) => {
+        console.log(err);
```
julian-londono, line 69: Same idea as above, helpful to preempt the error you print, like this:
https://github.com/cornell-dti/course-reviews-react-2.0/blob/f30c4ca2f22682e574e47f703c3fd1100d558f6b/server/methods.ts#L198

server/dbInit.ts
```diff
@@ -1,6 +1,165 @@
...
+          classTitle: cl.titleLong,
+          classFull: `${cl.subject.toUpperCase()} ${cl.catalogNbr} ${cl.titleLong}`,
+          classSems: [semester],
+          // TODO: provide initial values here?
+          classProfessors: null,
+          classRating: null,
+          classWorkload: null,
+          classDifficulty: null,
+        }).save().catch((err) => {
+          console.log(err);
```
julian-londono, line 117: Same idea as above, helpful to preempt the error you print, like this:

https://github.com/cornell-dti/course-reviews-react-2.0/blob/f30c4ca2f22682e574e47f703c3fd1100d558f6b/server/methods.ts#L198

julian-londono (changes requested):

server/dbInit.ts
khemritolya: This only fails in the case where something else has already failed, in which case we will have gotten another, more detailed error message ahead of time.

server/dbInit.ts
khemritolya: If this fails then the if at line 121 fails, and we get an error message about the particular class it was unable to insert.

server/dbInit.test.ts
khemritolya: If this fails, then tests start failing. In particular, one which checks if GORK 1110 is in the db. See line 106.

server/dbInit.ts
julian-londono: This is true, sounds good

server/dbInit.ts
julian-londono: Makes sense

server/dbInit.test.ts
julian-londono: Makes sense

julian-londono:
I agree with the functionality of `ALLOW_LOCAL`. One question: I assume that if `ALLOW_LOCAL=1` is set as an env_variable, this in-memory db is triggered by `yarn workspace server start` correct? If so, how does one know once this DB is populated and ready? I see [`console.log` in `server/server.ts`](https://github.com/cornell-dti/course-reviews-react-2.0/blob/052ebe1753bdbd9bc7678fdfdd24a2f5edb82809/server/server.ts#L31) but not later when it's finsihed

khemritolya:
The meteor-shim endpoints are created only after the DB has finished scraping and initializing the db: See [here](https://github.com/cornell-dti/course-reviews-react-2.0/blob/052ebe1753bdbd9bc7678fdfdd24a2f5edb82809/server/server.ts#L47)

julian-londono:
Awesome, just wanted to confirm - great solution and very useful for testing!

julian-londono (approved):

PR #198, opened 2020-10-06
Blank helpful

draycfarley:
### Summary 

This fixes the issue of empty helpful quantities on reviews.

### Test Plan 

Identified reviews locally with empty Helpful numbers. Then, I implemented my changes and verified that:
1. They now showed 0
2. Clicking over and over would update the number properly
3. The correct number was there after a refresh

Will test again on dev site using same criteria

### DIFF
--- a/client/src/ui/Review.jsx
+++ b/client/src/ui/Review.jsx
@@ -191,7 +191,7 @@ export default class Review extends Component {
                       alt={this.state.liked ? "Liked" : "Not Liked Yet"}
                     />
                     <p className="upvote-text">Helpful
-                        ({this.state.numLikes})</p>
+                        ({this.state.numLikes ? this.state.numLikes : 0})</p>
                   </button>
                 }
               </div>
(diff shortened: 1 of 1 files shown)


julian-londono (approved):
Works as intended on my local - confirmed values with React Dev Tools.

client/src/ui/Review.jsx
```diff
@@ -191,7 +199,7 @@ export default class Review extends Component {
                       alt={this.state.liked ? "Liked" : "Not Liked Yet"}
                     />
                     <p className="upvote-text">Helpful
-                        ({this.state.numLikes})</p>
+                        ({this.state.numLikes ? this.state.numLikes : 0})</p>
```
AkashAryal, line 202: I just recently found out about the "??" operator. a ?? b = b if a is null/undefined. Useful for things like this.

AkashAryal (approved):
Looks good. Tested locally and it seems to work as intended.

Issue #199, opened 2020-10-08
Overall Rating Bug


julian-londono:
Submitting a review on a class and approving it now causes the "Overall" rating of the class to be "NaN" and therefore show up as "-"

Issue currently exists in `master`

To recreate:
1. Find a class with existing reviews (and therefore existing "Overall", "Difficulty", "Workload" metrics on its class page)
2. Write a review
     - Note: this bug occurs with or without checking "Your experience was affected by COVID-19"
3. Approve this review
4. See that the class page for this class now has "Overall" as "-"
     - You can use React Dev Tools to see that the value of overall is "NaN"

Haven't tested that this is the case locally, but this is the case on the staging site:
https://cornellreviews-dev.herokuapp.com/

julian-londono:
Fixed by @khemritolya on #200 , closing

PR #200, opened 2020-10-10
Fix Issue #199

khemritolya:
### Summary 

See issue #199 . There was a problem were NaNs were introduced when computing gauge values. This should fix that problem, or at least it does for my local replication of the problem.

### Test Plan 

1. Create a server with a mongodb url pointing to dev db.
2. Create a review on a class with reviews which have NaNs. I used CS 2112, which failed before.
3. Approve Review
4. Confirm there are no problems

### Notes 

None

### Breaking Changes  

None

### DIFF
--- a/common/CourseCard.js
+++ b/common/CourseCard.js
@@ -75,36 +75,44 @@ function getGaugeValues(allReviews) {
   let sumWorkVirtual = 0;
 
   // create size counting variables
-  let countRatingAndDiffNormal = 0;
+  let countRatingNormal = 0;
+  let countDiffNormal = 0;
   let countWorkNormal = 0;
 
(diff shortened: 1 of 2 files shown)


draycfarley (approved):
Code changes look good and issue is fixed for me on local with this branch. There's a lot of unused code for handling weighting virtual reviews in CourseCard.js. Now that we have that implemented, let's talk about if we still want to do that on Tuesday.

julian-londono:
Can review by EOD today

julian-londono (approved):

julian-londono:
Solves locally for me as well.  Second the discussion Dray suggests we have on Tuesday.

PR #201, opened 2020-10-11
Add groundwork for end to end testing

draycfarley:
### Summary 

This pull request adds the testing framework Cypress to CU Reviews. Having end to end tests will make spotting bugs easier as we look to make the site more maintainable in the future.  Cypress is similar to Selenium in that it performs automated actions on a live browser, but Cypress was built specifically for testing.  In addition to adding this framework, this PR adds a test that will catch #199 if it occurs again in the future.

### Test Plan 

failTest4410

Tested using dev site but the tests can be switched to production or local in the future.

### Notes 

To run the tests locally, first add Cypress: `yarn workspace server add cypress`. 
Then run` yarn run cypress open` to open the Cypress app and run tests.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ db
 .meteor
 mongo/
 *error.log
+debug.log
--- a/cypress.json
+++ b/cypress.json
@@ -1,0 +1,1 @@
+{ "chromeWebSecurity": false}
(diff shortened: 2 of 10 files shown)


julian-londono (approved):
As you mentioned at the sync, it doesn't seem to work on Chrome even with the `{ "chromeWebSecurity": false}` option on.  Great start to end-to-end testing, looks great.

AkashAryal (approved):
Looks good. Since a lot of the tests depend on selectors, we should be careful when changing any frontend css/structure (especially when we start working on mobile). I had some trouble getting it to work with chrome or firefox but it works well on electron (which comes with cypress). Overall looks great.

Issue #203, opened 2020-10-14
Gauge circle "fullness" sticks when clicking a class with no reviews in search list


draycfarley:
The percentage a gauge is filled stays the same if you click from a class with reviews:
![image](https://user-images.githubusercontent.com/40501830/96028677-9d452580-0e27-11eb-9b73-0f3642789cc5.png)

to one with no reviews:
![image](https://user-images.githubusercontent.com/40501830/96028726-ac2bd800-0e27-11eb-9cae-bd5f301091fd.png)

PR #204, opened 2020-10-14
Reset gauge percentage when value is NaN

draycfarley:
### Summary 

This pull request fixes issue #203. It simply resets the percentage of the gauge to 0 when there is no metric available, where as the percentage would stay the same before.

### Test Plan 

Tested by clicking through classes and verifying values and percentage filled was changed appropriately. I also checked that the gauges still loaded properly on individual class pages.
Clicking on class with reviews:
image

And then clicking on one with no reviews right after:
image


julian-londono (approved):
Quick fix - much better user experience 🔥

PR #205, opened 2020-10-14
Remove virtual weight

draycfarley:
### Summary 

This pull request removes weighting of COVID affected/virtual reviews differently than other reviews. It also refactors the course card to get class metric values based on the values in the DB rather than recalculating them.

### Test Plan 

Tested by locally creating and then reporting/removing reviews for classes with reviews and without reviews to see if metrics changed as expected on the front end at each step.

### Notes 

common/CourseCard.js should probably be split up; I think it no longer contains functions used by both the back end and the front end.

### Breaking Changes  

Changing how the front end gets the values for gauges has potential to cause errors but looks good from my testing.


julian-londono (approved):
Code looks clean to me.  Tested locally and new metrics work for me as well.

PR #207, opened 2020-10-26
Create example pure express.js endpoint

khemritolya:
### Summary

This pull request is the first step towards deprecating the Meteor Shim on the backend

This PR:

* Creates an example endpoint to pattern other after
* Sets up a small framework which abstracts away boiler-plate code
* Sets up testing for the newly created endpoints

This is a draft pull request while we decide if this is a good pattern, and if there are any things about the pattern that we want to change.

### Test Plan 

Write tests for the endpoints. See endpoints.tests.ts

### Notes 

TODO: Discuss the merits.

### Breaking Changes 

None. Deprecation will come later.

### DIFF
--- a/server/endpoints.test.ts
+++ b/server/endpoints.test.ts
@@ -1,0 +1,92 @@
+import mongoose from 'mongoose';
+import { MongoMemoryServer } from 'mongodb-memory-server';
+import express from "express";
+
+import axios from 'axios';
+import { configure } from "./endpoints";
+import { Classes, Students, Subjects } from './dbDefs';
+
+let mongoServer: MongoMemoryServer;
+
+const testingPort = 17760;
+
+beforeAll(async () => {
+  // get mongoose all set up
+  mongoServer = new MongoMemoryServer();
+  const mongoUri = await mongoServer.getUri();
+  await mongoose.connect(mongoUri, { useNewUrlParser: true, useUnifiedTopology: true });
+
+  await mongoose.connection.collections.classes.createIndex({ classFull: "text" });
+  await mongoose.connection.collections.subjects.createIndex({ subShort: "text" });
+  await mongoose.connection.collections.professors.createIndex({ fullName: "text" });
+
+  const newUser = new Students({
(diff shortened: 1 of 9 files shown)


server/endpoints.ts
```diff
@@ -0,0 +1,116 @@
...
+ */
+export function configure(app: express.Application) {
+  const methods = express.Router();
+  methods.use(express.json());
+  app.use(express.json());
+
+  register(app, "getClassesByQuery", getClassesByQuery);
+}
+
+function register<T>(app: express.Application, name: string, endpoint: Endpoint<T>) {
```
draycfarley, line 23: Maybe this has to do with the shim or using Typescript with Express, but why not use app.use("/v2")? When I've used Express without TS that was sufficient for directing requests to a set of endpoints.

khemritolya: I think that this is more explicit about where all the new endpoints are listening. We can consider changing it though.

server/endpoints.test.ts
```diff
@@ -0,0 +1,92 @@
...
+
+import axios from 'axios';
+import { configure } from "./endpoints";
+import { Classes, Students, Subjects } from './dbDefs';
+
+let mongoServer: MongoMemoryServer;
+
+const testingPort = 17760;
+
+beforeAll(async () => {
```
AkashAryal, line 13: In the future we should probably set up a jest global setup and global teardown so we don't have to repeat this every test file.

khemritolya: I agree in the long run. However, in the short run, all tests modify the DB, and they do so in different ways. For tests to be deterministic, this is a necessity. Again, maybe in the future we can be smarter about this.

julian-londono (changes requested):
Everything looks good to me!  Could you create an `/endpoints/` directory where we could house back end functions by components in their own folders as discussed during the sync?

E.g. `getClassesByQuery` would go in `SearchBar.ts` because it's used by this component

julian-londono (approved):

PR #208, opened 2020-10-30
Front-End Scraper Interface

julian-londono:
### Summary 

This PR builds upon #195 by adding the front-end interface to update semesters from Course Roster data.

- [x] Loading text on front-end while semester is being updated (scraping and processing data
- [x] Front-end confirmation that the scraping was successfully completed or errored

### Test Plan 

Tested locally - updating a semester takes 1-2 minutes.

Will test on PR dev server

### Notes 

This **does not link cross-listed courses**.  This was something that #195 did not implement yet and therefore something we will add to our backlog.  There is an old `addCrossList` function but we are currently unsure on its dependability and are therefore holding off on this functionality in exchange for getting this functionality out and on prod.

### DIFF
--- a/client/src/meteor-shim.ts
+++ b/client/src/meteor-shim.ts
@@ -16,7 +16,7 @@ export class MeteorShim {
     this._subscribe(name, args.slice(0, -1), args[args.length - 1]);
   }
 
-  setup() { this._axios = axios.create({ baseURL: '', timeout: 30000 }); }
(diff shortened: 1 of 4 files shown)


draycfarley (commented):
Since this PR isn't urgently needed in master, I think it could be better to have our base for Express endpoints set up and then use that instead of meteor methods. It could also be a helpful way to see if any unexpected errors come up when using them without jeopardizing user facing features.

julian-londono:
Absolute agree, closing this for now and will re-open once it's converted to Express

PR #209, opened 2020-11-01
Liking a review test and searching tests

draycfarley:
### Summary 

This pull request adds onto the end to end testing previously implemented. Tests added:
- Liking a review and checking the like number stays updated after refreshing the page
- Searching for a class by course subject and number exactly
- Searching for  a class by keywords in its title, i.e. "Functional Programming"
### Test Plan 

Ran on Cypress electron browser locally using the dev site for the test functions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ dist
 build
 mongo_backups/*
 !mongo_backups/.gitkeep
+mongodb-binaries/
 .git
 .vscode
 .env
@@ -13,3 +14,4 @@ db
 .meteor
 mongo/
 *error.log
+debug.log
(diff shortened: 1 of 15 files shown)


julian-londono:
Don't mean to keep this open for too long.  Is the only way around the authentication tests to log-in an Edge locally before running the Cypress tests?

draycfarley:
> Don't mean to keep this open for too long. Is the only way around the authentication tests to log-in an Edge locally before running the Cypress tests?

As the app is currently structured, yes which is a missed insight on my part. I think the best way around this would be to perhaps refactor our code to un-require admin authentication when running the tests on a PR check.

julian-londono:
Gotcha, I'd say that's beyond the scope of this PR anyway then.  Let's merge and eventually see how we can include this on the GitHub CI on a PR check?

julian-londono (approved):

PR #210, opened 2020-11-04
Express Endpoint for viewing Reviews

AkashAryal:
### Summary 

This pull request creates express endpoints for viewing reviews. Currently we have two meteor methods that do this: 
`getReviewsByCourseId` and `getReviewsByProfessor`. Doing a quick search through the project shows that `getReviewsByProfessor` is not used anywhere, so this PR will have `getReviewsByCourseId` and `getCourseById` (used in `getReviewsByCourseId`). Other endpoints for reviews that change data ex: makeVisible, reportReview are not included in this PR for viewing reviews.

### Test Plan 
Added tests in `server/endpoints/Review.test.ts`

### Notes 
Currently it seems like we are organizing endpoints by where they are used. EX: `getClassesByQuery` is used in `SearchBar` so we put it in `SearchBar.ts`. Looking through the meteor methods we have, however, I don't think such a method would work very well as many functions are used in multiple places. I think it would be better if we split endpoints by their jobs. For example, all endpoints that have to do with getting/updating reviews would go in a file `Review.ts` and endpoints dealing with auth ex: `tokenIsAdmin` would go in `Auth.ts` and so on.

### DIFF
--- a/server/endpoints/Review.test.ts
+++ b/server/endpoints/Review.test.ts
@@ -3,13 +3,14 @@ import { MongoMemoryServer } from 'mongodb-memory-server';
 import express from "express";
 
 import axios from 'axios';
+import { Review } from 'common';
 import { configure } from "../endpoints";
-import { Classes, Reviews } from "../dbDefs";
+import { Classes, Students, Subjects, Reviews } from "../dbDefs";
 
 let mongoServer: MongoMemoryServer;
 let serverCloseHandle;
 
-const testingPort = 8000;
+const testingPort = 37760;
 
 beforeAll(async () => {
   // get mongoose all set up
@@ -74,11 +75,11 @@ beforeAll(async () => {
 afterAll(async () => {
   await mongoose.disconnect();
   await mongoServer.stop();
-  await serverCloseHandle.close();
+  serverCloseHandle.close();
 });
 
 describe('tests', () => {
-  it('getReviewsByCourseId - posting review', async () => {
+  it('getClassesByQuery-works', async () => {
     const res = await axios.post(`http://localhost:${testingPort}/v2/getReviewsByCourseId`, { courseId: "oH37S3mJ4eAsktypy" });
     expect(res.data.result.length).toBe(2);
   });
--- a/server/endpoints/Review.test.ts
+++ b/server/endpoints/Review.test.ts
@@ -9,7 +9,7 @@ import { Classes, Reviews } from "../dbDefs";
 let mongoServer: MongoMemoryServer;
 let serverCloseHandle;
 
-const testingPort = 37760;
+const testingPort = 8000;
 
 beforeAll(async () => {
   // get mongoose all set up
@@ -62,9 +62,9 @@ beforeAll(async () => {
     reported: 0,
   });
 
-  await review2.save();
-  await review1.save();
-  await c2110.save();
+  review2.save();
+  review1.save();
+  c2110.save();
   // Set up a mock version of the v2 endpoints to test against
   const app = express();
   serverCloseHandle = app.listen(testingPort, async () => { });
@@ -74,11 +74,11 @@ beforeAll(async () => {
 afterAll(async () => {
   await mongoose.disconnect();
   await mongoServer.stop();
-  serverCloseHandle.close();
+  await serverCloseHandle.close();
 });
 
 describe('tests', () => {
-  it('getClassesByQuery-works', async () => {
+  it('getReviewsByCourseId - posting review', async () => {
     const res = await axios.post(`http://localhost:${testingPort}/v2/getReviewsByCourseId`, { courseId: "oH37S3mJ4eAsktypy" });
     expect(res.data.result.length).toBe(2);
   });
--- a/server/endpoints/Review.test.ts
+++ b/server/endpoints/Review.test.ts
@@ -3,14 +3,13 @@ import { MongoMemoryServer } from 'mongodb-memory-server';
 import express from "express";
 
 import axios from 'axios';
-import { Review } from 'common';
 import { configure } from "../endpoints";
-import { Classes, Students, Subjects, Reviews } from "../dbDefs";
+import { Classes, Reviews } from "../dbDefs";
 
 let mongoServer: MongoMemoryServer;
 let serverCloseHandle;
 
-const testingPort = 37760;
+const testingPort = 8000;
 
 beforeAll(async () => {
   // get mongoose all set up
@@ -75,11 +74,11 @@ beforeAll(async () => {
 afterAll(async () => {
   await mongoose.disconnect();
   await mongoServer.stop();
-  serverCloseHandle.close();
+  await serverCloseHandle.close();
 });
 
 describe('tests', () => {
-  it('getClassesByQuery-works', async () => {
+  it('getReviewsByCourseId - posting review', async () => {
     const res = await axios.post(`http://localhost:${testingPort}/v2/getReviewsByCourseId`, { courseId: "oH37S3mJ4eAsktypy" });
     expect(res.data.result.length).toBe(2);
   });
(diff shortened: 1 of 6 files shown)


server/endpoints/Review.test.ts
```diff
@@ -0,0 +1,85 @@
...
+    quality: 5,
+    class: "oH37S3mJ4eAsktypy",
+    grade: 6,
+    date: new Date().toISOString(),
+    atten: 0,
+    visible: 1,
+    reported: 0,
+  });
+
+  review2.save();
```
khemritolya, line 65: This returns a Promise. Although it doesn't seem to cause any issues, it should probably still be awaited.

khemritolya: Same for the next two lines

khemritolya (commented):
I think that much of this is good, there are just some minor things to change.

I agree with your comment in "Notes" about the names for the various files. Tenatively, I suggest that we rename SearchBar to just Search (I can do this in my PR on finishing the rest of search). As for other files, I think there should be seperate a file for anything on the Admin Page. As for inserting reviews, I think there should be another seperate file for user interaction (inserting, liking, etc.) which require auth, and Reviews.ts should handle only "publicly" available data. But let me know what you think @AkashAryal @julian-londono @draycfarley

server/endpoints/Review.ts
```diff
@@ -0,0 +1,54 @@
+import { body } from "express-validator";
+import { getCrossListOR } from "common/CourseCard";
+import { Endpoint } from "../endpoints";
+import { Reviews, Classes } from "../dbDefs";
+
+interface CourseId {
+  courseId: string;
+}
+
+export const getCourseById: Endpoint<CourseId> = {
```
khemritolya, line 10: It would be nice to have a little bit of documentation here. Just a couple of words should do, I think.

server/endpoints/Review.ts
```diff
@@ -0,0 +1,54 @@
...
+      // eslint-disable-next-line no-console
+      console.log("Error: at 'getCourseById' method");
+      // eslint-disable-next-line no-console
+      console.log(error);
+      return null;
+    }
+  },
+};
+
+// eslint-disable-next-line import/prefer-default-export
```
khemritolya, line 30: It would be nice to have a little bit of documentation here. Just a couple of words should do, I think.

server/endpoints/Review.test.ts
```diff
@@ -0,0 +1,85 @@
...
+afterAll(async () => {
+  await mongoose.disconnect();
+  await mongoServer.stop();
+  await serverCloseHandle.close();
+});
+
+describe('tests', () => {
+  it('getReviewsByCourseId - posting review', async () => {
+    const res = await axios.post(`http://localhost:${testingPort}/v2/getReviewsByCourseId`, { courseId: "oH37S3mJ4eAsktypy" });
+    expect(res.data.result.length).toBe(2);
```
julian-londono, line 83: I think this test could be a bit more comprehensive.  Something sufficient could be after checking the length, confirm that each review object's `class` id matches this `oH37S3mJ4eAsktypy `

julian-londono (changes requested):

server/endpoints/Review.test.ts
julian-londono: This is fixed, resolving

server/endpoints/Review.ts
AkashAryal: I copied the comments from the original function but i think you are right. Not sure why "local" was used. I can remove it.

server/endpoints/Review.test.ts
khemritolya: I second this concern. Given that ordering is not guaranteed from mongoose, you could map the results to their id (or any other unique field like class number) using .map on the returned list, and the use the .contains() method that jest provides to verify that all the values you want are in list.

server/endpoints/Review.ts
khemritolya: You are correct Julian, "Local" is wrong in this context. Generally speaking it should be pulling from staging/prod, which is not local.

server/endpoints/Review.ts
```diff
@@ -0,0 +1,60 @@
...
+      if (regex.test(courseId.courseId)) {
+        return await Classes.findOne({ _id: courseId.courseId }).exec();
+      }
+      return null;
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.log("Error: at 'getCourseById' method");
+      // eslint-disable-next-line no-console
+      console.log(error);
+      return null;
```
khemritolya, line 29: I would return some sort of error message for these nulls. See how I did it in the example in Search.ts: e.g ```return { error: "Malformed Query" };```

This goes for several other places where null is returned. Let me know if you disagree though.

server/endpoints/Review.test.ts
```diff
@@ -0,0 +1,85 @@
...
+import express from "express";
+
+import axios from 'axios';
+import { configure } from "../endpoints";
+import { Classes, Reviews } from "../dbDefs";
+
+let mongoServer: MongoMemoryServer;
+let serverCloseHandle;
+
+const testingPort = 8000;
```
khemritolya, line 12: 8000 works, but we may want to put it on different ports to avoid any potential conficts with other tests in the future. I generally try to go for >10k.

This is not a big concern though.

khemritolya (commented):
Generally good, there are some last minor concerns I still have however. Let me know if you disagree with any of my assesments. 

Also @julian-londono let me know if I clarified that "local" thing sufficiently.

server/endpoints/Review.ts
julian-londono: Agree here, error descriptions are generally helpful

khemritolya (approved):
Looks good to me! Though you need to merge in some changes from master.

julian-londono (approved):

julian-londono:
All good by me as well, just make sure to merge/rebase `master` and check that your tests all pass

server/endpoints/Review.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Get list of review objects for given class from class _id
+ */
+export const getReviewsByCourseId: Endpoint<CourseId> = {
+  guard: [body("courseId").notEmpty().isAscii()],
+  callback: async (courseId: CourseId) => {
+    try {
+      const regex = new RegExp(/^(?=.*[A-Z])/i);
+      if (regex.test(courseId.courseId)) {
+        const course = await getCourseById.callback(courseId);
```
draycfarley, line 42: Sorry for the late comment but could the common functionality between this endpoint and getCourseById be factored out into a function called by both? Thinking through future endpoints, they'll probably need to get courses by ids as well. While we can simply call the endpoint again like is done here, it isn't best practice: https://softwareengineering.stackexchange.com/questions/345406/is-it-bad-design-to-internally-call-api-endpoints-from-within-the-api-instance.

draycfarley (changes requested):

server/endpoints/Review.ts
AkashAryal: Yea that makes sense. I can remove getCourseById from the endpoints just have that be a utility function. Should I make a utils.ts and put it there (since its not an endpoint it doesn't make sense to keep it in Review.ts)? Or did you have something else in mind?

draycfarley: I think you can keep the getCourseById endpoint because the front end uses the getCourseById Meteor method in a few places and put a function that actually gets the course in that utils file you suggested. Then the getCourseById and any other endpoints that need to get course info can call that function

AkashAryal: Ah, I see what you mean. I think I misunderstood the first time.

draycfarley (approved):
Looks good

PR #211, opened 2020-11-08
Express endpoints for admin authentication

draycfarley:
### Summary 

This pull request implements an express endpoint for checking if a token is for an admin user or not.

### Test Plan 
Tested through Postman with my own Google account token as well as through adding unit tests which mock the Google client's response in Auth.test.ts.

image

### Notes 

We may want to change the test file to mock the Google client object instead of the getVerificationTicket in the future. I would like others' thoughts on this as well.


server/endpoints/Auth.ts
```diff
@@ -0,0 +1,110 @@
...
+/**
+   * Returns true if [netid] matches the netid in the email of the JSON
+   * web token. False otherwise.
+   * This method authenticates the user token through the Google API.
+   * @param token: google auth token
+   * @param netid: netid to verify
+   * @requires that you have a handleVerifyError, like as follows:
+   * verify(token, function(){//do whatever}).catch(function(error){
+   * handleVerifyError(error, res);
+   */
```
julian-londono, line 22: I don't think this function actually requires this anymore - seems like old documentation

server/endpoints/Auth.ts
```diff
@@ -0,0 +1,110 @@
...
+  try {
+    if (token === null) {
+      console.log("Token was undefined in getVerificationTicket");
+      return null; // Token was undefined
+    }
+    const ticket = await client.verifyIdToken({
+      idToken: token,
+      audience: "836283700372-msku5vqaolmgvh3q1nvcqm3d6cgiu0v1.apps.googleusercontent.com", // Specify the CLIENT_ID of the app that accesses the backend
+      // Or, if multiple clients access the backend:
+      // [CLIENT_ID_1, CLIENT_ID_2, CLIENT_ID_3]
```
julian-londono, line 33: These comments about `CLIENT_ID` stuff is old boilerplate comments from a tutorial -  can be removed

server/endpoints/Auth.ts
```diff
@@ -0,0 +1,110 @@
...
+   * Used in the .catch when verify is used, handles whatever should be done
+   * @param errorObj (required) the error that is returned from the .catch
+   * @param res the response object
+   * @return {boolean} true if their token is too old, false if some other error
+   * @requires that you have the verify function, like as follows:
+   * verify(token, function(){//do whatever}).catch(function(error){
+   *        handleVerifyError(error, res);
+   * }
+   */
+export const handleVerifyError = (errorObj, res) => {
```
julian-londono, line 54: Related to comment above - I don't think `handleVerifyError` is ever explicitly used anywhere.  Could somehow be implicitly called but if not it should be removed

server/endpoints/Auth.ts
```diff
@@ -0,0 +1,110 @@
...
+  }
+  return false;
+};
+
+// Get a user with this netId from the Users collection in the local database
+export const getUserByNetId = async (netId: string) => {
+  try {
+    const regex = new RegExp(/^(?=.*[A-Z0-9])/i);
+    if (regex.test(netId)) {
+      return await Students.findOne({ netId }).exec();
```
julian-londono, line 72: Not 100% positive our `_id` field in `Student` and the following are equivalent but the [Mongoose docs seem to recommend using `findById()`](https://mongoosejs.com/docs/api.html#model_Model.findOne) in this case

julian-londono (changes requested):

julian-londono:
### re: Thoughts on mocking the Google client object

I think this would be useful in the future as it would therefore add tests to `getVerificationTicket` itself.  For how, however, I think the tests set up so far are good.  I'd save that for a future PR.

server/endpoints/Auth.ts
draycfarley: The netID is different from the document ID

draycfarley:
@julian-londono Could you mark the changes as resolved?

julian-londono (approved):

PR #213, opened 2020-12-01
Express endpoints for admin authenticated endpoints

draycfarley:
### Summary 

This pull request creates endpoints for the 5 Meteor methods that require admin authentication to carry out that are not stats dashboard related. It also adds unit tests where applicable and refactors Auth.ts to account for the new utils.ts folder.

### Test Plan 

Added unit tests in AdminActions.test.ts

### Notes 

Unit tests weren't added for the two professor functions because their functionality falls under behavior that should be tested in dbInit.test.ts. There aren't any unit tests for the underlying dbInit functions used already. I don't think there's much use in unit tests for those endpoints/functions, but I wanted to note in case it raised concerns.

### DIFF
--- a/server/endpoints/AdminActions.ts
+++ b/server/endpoints/AdminActions.ts
@@ -3,8 +3,9 @@ import { body } from "express-validator";
 import { Endpoint } from "../endpoints";
 import { Reviews, ReviewDocument, Classes } from "../dbDefs";
 import { verifyToken } from "./Auth";
-import { updateProfessors, findAllSemesters } from "../dbInit"
+import { updateProfessors, findAllSemesters, resetProfessorArray } from "../dbInit"
 import { getCrossListOR, getMetricValues } from "../../common/CourseCard";
+import { getCourseById } from "./utils";
 
 // The type for a request with an admin action for a review
 interface AdminReviewRequest {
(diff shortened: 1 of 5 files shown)


server/endpoints/AdminActions.ts
```diff
@@ -0,0 +1,189 @@
...
+ * Approve a submitted review and make it visible
+ */
+export const makeReviewVisible: Endpoint<AdminReviewRequest> = {
+  guard: [body("review").notEmpty(), body("token").notEmpty().isAscii()],
+  callback: async (adminReviewRequest: AdminReviewRequest) => {
+    try {
+      // check: make sure review id is valid and non-malicious
+      const userIsAdmin = await verifyToken(adminReviewRequest.token);
+      const regex = new RegExp(/^(?=.*[A-Z0-9])/i);
+      if (regex.test(adminReviewRequest.review._id) && userIsAdmin) {
```
julian-londono, line 64: Something @khemritolya demonstrated is that we can replace these regex checks by doing that check in the guard: [example from #214 ](https://github.com/cornell-dti/course-reviews-react-2.0/pull/214/files#diff-1d08deae3ced372c187d247f952b042240ffa819c56f331fe8d9760c26abba27R48-R56).

Therefore in this case I'd recommend replacing the regex check with this in the guard chain:
`body("review._id").isAlphanumeric()`

server/endpoints/AdminActions.ts
```diff
@@ -0,0 +1,189 @@
...
+ * "Undo" the reporting of a flagged review and make it visible again
+ */
+export const undoReportReview: Endpoint<AdminReviewRequest> = {
+  guard: [body("review").notEmpty(), body("token").notEmpty().isAscii()],
+  callback: async (adminReviewRequest: AdminReviewRequest) => {
+    try {
+      const userIsAdmin = await verifyToken(adminReviewRequest.token);
+      // check: make sure review id is valid and non-malicious
+      const regex = new RegExp(/^(?=.*[A-Z0-9])/i);
+      if (regex.test(adminReviewRequest.review._id) && userIsAdmin) {
```
julian-londono, line 89: Same regex comment as above

julian-londono (changes requested):

julian-londono:
For reference, an entire list of the validations that come with `express-validator` is here: https://github.com/validatorjs/validator.js#validators

server/endpoints/AdminActions.ts
draycfarley: Noted, done 👍

julian-londono (approved):

PR #214, opened 2020-12-01
Odds and Ends, and wrapping up new public API

khemritolya:
### Summary 

This PR is another part of the long road to severing our dependence on the Meteor shim. It also tries to get rid of some of the bad code that is infiltrating our codebase as holdovers from methods.ts.

In particular, it added another method which (as far as I was able to tell) is the final one used for public accesses (i.e. not cornell student submission dependent). Let me know if there was one I missed.

Additionally, this PR removes the obsessive regex checking in some of the backend method. It adds nothing that is not provided in a better way by the guard pattern and express-validator. See the test in Reviews.test.ts which show that these features really do work as intended. Going forward, I think it makes sense to remove these regex checks if they can be replaced by equivalent code in the guard.

### Test Plan 

See the additional unit tests in Reviews.test.ts.

### Notes 

 Also see some awaits in Search.test.ts which were missing, and theoretically might have caused problems down the line. I also refractored a name for an interface.

### Breaking Changes 

None yet, as per usual.

### DIFF
--- a/server/endpoints.ts
+++ b/server/endpoints.ts
@@ -1,6 +1,6 @@
 import express from "express";
 import { validationResult, ValidationChain } from "express-validator";
-import { getReviewsByCourseId, getCourseById } from "./endpoints/Review";
+import { getReviewsByCourseId, getCourseById, getCourseByInfo } from "./endpoints/Review";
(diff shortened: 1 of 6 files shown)


draycfarley (approved):
Looks good to me

PR #215, opened 2020-12-03
Express endpoints for inserting reviews and Users

AkashAryal:
### Summary 
This pull request adapts the endpoint for inserting reviews. It turns out that inserting reviews requires inserting users so this PR adapts that as well. 

### Test Plan 
Tests are in Review.test.ts. Tested inserting review and user by asserting that the new user and review are in the db.

### Notes 
Currently I have put the endpoint for inserting users in Review.ts but we can move it when we decide on a structure.

### DIFF
--- a/server/endpoints/Review.ts
+++ b/server/endpoints/Review.ts
@@ -1,4 +1,4 @@
-import { body, ValidationChain } from "express-validator";
+import { body } from "express-validator";
 import { getCrossListOR } from "common/CourseCard";
 import { Review } from "common";
 import { includesProfanity } from "common/profanity";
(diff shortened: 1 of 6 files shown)


server/endpoints/Review.ts
```diff
@@ -1,14 +1,29 @@
-import { body } from "express-validator";
+import { body, ValidationChain } from "express-validator";
```
julian-londono, line 1: nit: `ValidationChain` seems unused, unnecessary import in this file

julian-londono (approved):

server/endpoints/utils.ts
```diff
@@ -18,3 +22,56 @@ export const getCourseById = async (courseId: CourseIdQuery) => {
...
+    console.log(error);
+    return 0;
+  }
+};
+
+/**
+ * Creates a ValidationChain[] where the json object denoted by [jsonFieldName]
+ * has non-empty fields listed in [fields].
+ */
+export const JSONNonempty = (jsonFieldName: string, fields: string[]) => {
```
julian-londono, line 71: Great util function and use in `Review.ts`!

PR #219, opened 2020-12-08
Client side uses search express

draycfarley:
### Summary 

This pull request replaces Meteor calls for getting classes by keyword/query on the front end. It also moves the getClassesByQuery Meteor method that uses regular expressions to the getClassesByQuery endpoint and moved its regular expression check to the endpoint's guard. 

### Test Plan 

- checked existing unit tests worked for getClassesByQuery endpoint
- Tested searching from home page, results page, course page

### Notes 

We should do thorough testing on the dev site again since searching is so essential to using the site.

### Breaking Changes  

N/A

### DIFF
--- a/client/src/ui/SearchBar.jsx
+++ b/client/src/ui/SearchBar.jsx
@@ -86,21 +86,12 @@ export default class SearchBar extends Component {
           });
         }
         else {
-          Meteor.call("getClassesByQuery", this.state.query, (err, queryCourseList) => {
-            if (!err && queryCourseList && queryCourseList.length !== 0) {
-              // Save the list of Class objects that matches the request
-              this.setState({
-                allCourses: queryCourseList
-              });
-            }
-            else {
               this.setState({
                 allCourses: []
               });
             }
-          })
         }
-      })
+      )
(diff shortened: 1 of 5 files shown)


client/src/ui/SearchBar.jsx
```diff
@@ -112,21 +107,23 @@ export default class SearchBar extends Component {
...
 
-      Meteor.call("getProfessorsByName", this.state.query, (err, subjectList) => {
-        if (!err && subjectList && subjectList.length !== 0) {
+      axios.post(`/v2/getProfessorsByQuery`, { query: this.state.query }).then(response => {
+        const professorList = response.data.result;
+        if (professorList && professorList.length !== 0) {
           // Save the list of Subject objects that matches the request
           this.setState({
-            allProfessors: subjectList
+            allProfessors: professorList
```
julian-londono, line 117: Good catch!

julian-londono (approved):
Works for me locally:
- Searched for exact match
- Searched by professor
- Searched by title

I agree we'll thoroughly test on the dev site.

PR #220, opened 2020-12-08
Admin live behavior

draycfarley:
### Summary 

Changes Admin review page to display reviews based on state, not props, and uses this to remove reviews from view after they've been approved or removed.

### Test Plan 

Tested locally through approving, removing, and unreporting reviews on the admin page.

### Notes 

Took a long time to update when running locally but I believe this was due to the backend call taking a long time.

### Breaking Changes  

N/A


julian-londono (approved):
Works (and fast) locally for me!

PR #221, opened 2020-12-11
Add Cypress tests to CI workflow

draycfarley:
### Summary 

- adds an environment variable check to see if admin authentication is turned off
- adds running Cypress checks to the CI pipeline

### Test Plan 

Unfortunately I couldn't find a good way to test locally on my machine, so it seems that this PR itself is the test. I'll add commits to tweak and get the changes to the workflow working as expected.

### Notes 

After we have Cypress set up, we'll need to keep in mind that front end changes could break the tests in the future and we'll have to change them.


.github/workflows/ci-workflow.yml
```diff
@@ -17,5 +17,15 @@ jobs:
...
+      - name: Unit Tests
         run: yarn workspace server test
+      - name: End to End Tests
+        uses: cypress-io/github-action@v1
+        with:
+          start: yarn workspace server start && yarn workspace client start
+          wait-on: http://localhost:3000
+          wait-on-timeout: 180
+        env: 
+          ADMIN_DISABLED: "1"
```
julian-londono, line 29: Good call with this `env variable`

draycfarley:
Looks like it's working! 🥳 Link to documentation about how the GitHub side works and what changes were needed on the backend/frontend to make it work: https://cornelldti.slab.com/posts/cu-reviews-end-to-end-test-workflow-0l1p43rg

khemritolya:
Unfortunately, this PR has too many issues for us to merge at this time.
1. It doesn't run on my machine, or Akash's
2. The mobile redesign has significantly impacted much of the frontend
3. It causes a huge number of linter warnings

In the future, we may consider re-introducing end-to-end testing, however, at this time, we don't have the ability to do so.

Issue #218, opened 2020-12-08
Huge decimal for CS 2110


draycfarley:
Noticed this issue on dev and prod where CS 2110 has a long decimal value displayed in results:
![image](https://user-images.githubusercontent.com/40501830/101515401-50279080-394c-11eb-9549-69a423361d4d.png)

PR #223, opened 2020-12-12
Fixed rounding issue

draycfarley:
### Summary 

This fixes issue #218 where CS 2110 had a huuuge decimal. It also applies the same fix to gauge values in case that comes up in the future.

image

### Test Plan 

Verified that classes' metrics from results and course page still looked good, both after searching and submitting reviews.

### DIFF
--- a/client/src/ui/FilteredResult.tsx
+++ b/client/src/ui/FilteredResult.tsx
@@ -42,8 +42,8 @@ export default class FilteredResult extends Component<Props,State> {
     };
 
     this.getColor = this.getColor.bind(this);
-    this.updateSortBy = this.updateSortBy.bind(this);
-    this.updateSortByTitle = this.updateSortByTitle.bind(this);
+    this.getSortNumber = this.getSortNumber.bind(this);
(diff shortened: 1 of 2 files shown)


client/src/ui/FilteredResult.tsx
```diff
@@ -128,7 +128,7 @@ export default class FilteredResult extends Component<Props,State> {
               <strong>{this.updateSortByTitle()}</strong>
             </p>
             <p className="result-card-sort-by-value">
-              {this.updateSortBy()}
+              {isNaN(Number(this.updateSortBy())) ? this.updateSortBy() : Number(this.updateSortBy()).toFixed(1)}
```
julian-londono, line 131: Two thoughts:
- I think we should change the name of `updateSortBy` to something more like `getSortNumber`
- I think the logic for rounding should be an optional parameter to this function
  - E.g. `getSortNumber(1)` for getting number rounded to 1 decimal; `getSortNumber()` for no rounding

julian-londono (approved):
Sweet thanks for cleaning this up!

PR #225, opened 2020-12-13
Remove defunct popup code

draycfarley:
### Summary 

The "Leave a review" pop up was a great idea but unfortunately fell victim to super complicated state management and messy handler calls. Since the code hasn't been touched in a year, was done for the old design of the site, and is pervasive in several components, I decided to remove it as part of code clean up day. The old code can be accessed if needed from prior commits but the whole thing should probably be refactored.

### Test Plan 

Tested submitting and approving reviews to make sure the process still worked and looked like normal.

### Notes 

We'll want to again test on dev site since several key components were touched in this PR.

### Breaking Changes  

N/A


julian-londono (approved):
Huge refactor! A shame that the popup was short-lived and turned messy but definitely the right choice to refactor it away.

Everything seems to work locally.

PR #227, opened 2020-12-17
New like/dislike functionality

khemritolya:
### Summary 

This migrates over the liking funcitonality to the new framework. It also "fixes" a security vulnerability where you could increment the like/dislike arbitrarily with a little bit of skill. This has been mostly fixed. This will also be revisited very early next semester.

### Test Plan 

See the unit tests in Review.test.ts

Tested on local with the method that previously worked to increment arbitrarily. Method does not work anymore. Recommend to test on local to verify it all works for you.

### Notes 

This will need to be revised early next semester when we introduce profiles on the backend.

### Breaking Changes  

Removes old like/dislike functionality


AkashAryal:
@khemritolya could you merge master in?

server/methods.ts
```diff
@@ -191,51 +191,17 @@ Meteor.methods({
     }
   },
   /**
-   * Increment the number of likes a review has gotten by 1.
-   *
-   * Returns 1 on success
-   * Returns 0 on error
+   * Deprecated for security reasons
    */
   async incrementLike(id) {
```
AkashAryal, line 196: I don't think this is used anywhere (since you updated the front end with axios calls in this PR too) so we can remove this function (unless you want to keep it for book keeping or something).

server/methods.ts
```diff
@@ -191,51 +191,17 @@ Meteor.methods({
...
   },
 
   /**
-   * Decrement the number of likes a review has gotten by 1.
-   *
-   * Returns 1 on success
-   * Returns 0 on error
+   * Deprecated for security reasons
    */
   async decrementLike(id) {
```
AkashAryal, line 203: Same here

khemritolya:
Done! @AkashAryal let me know if there is anything else that I should fix!

AkashAryal:
@khemritolya If I like then dislike then try to like again, it does not work. 

Also I might be understanding it wrong but if person A likes then person B with a different ip likes doesn't that mean person A can like again?

AkashAryal (approved):

PR #228, opened 2020-12-17
Adminchart endpoints

AkashAryal:
### Summary 

This pull request adds methods required for the admin chart

### Test Plan 

Test cases are in AdminChart.test.ts. We test by asserting that the various stats the function get are correct.


khemritolya (approved):
Looks good! Some of the methods are a bit hard to read -- though I know they are literal migrations of the old. We should think about cleaning them up some more when we integrate with the frontend.

PR #236, opened 2021-02-23
Changed frontend to use axios for stats component functions

AkashAryal:
### Summary 

This pull request updates the front end to use axios to call the new express endpoints for functions **totalReviews**, **howManyReviewsEachClass**, **howManyEachClass**, and **topSubjects**. These are the main non chart functions in the admin page.

### Test Plan 
Checked locations where these functions were used to ensure things still work.


server/endpoints/AdminChart.ts
```diff
@@ -187,7 +187,7 @@ const topSubjectsCB = async (_ctx: Context) => {
  * Returns the top 15 subjects (in terms of number of reviews)
  */
 export const topSubjects: Endpoint<unknown> = {
-  guard: [body("token").notEmpty()],
+  guard: [],
```
khemritolya, line 190: Why did you remove the guard?

AkashAryal: When I first migrated the backend topSubjects method I thought it did take in an auth token (thus i put it in the guard) but it turns out that it does not take in anything (so i removed it in this PR). Should we alter the topSubjects method to require a token? (I think that makes sense too).

khemritolya: If we don't already, we should probably gate all the admin functions behind tokens.

khemritolya (commented):
Looks good, but I tested a bit more, and I found that there seem to still be calls to the old topSubjects around, at least when I check console output. Can you confirm? Other ones seem to sucessfully be not longer called.

AkashAryal:
> Looks good, but I tested a bit more, and I found that there seem to still be calls to the old topSubjects around, at least when I check console output. Can you confirm? Other ones seem to sucessfully be not longer called.

The method calling the old topSubjects is the big chart method. Once I update the chart function in the next pr, the issue should fix itself.

khemritolya (approved):
Looks good!

PR #237, opened 2021-02-27
Update all course metrics et. al.

khemritolya:
### Summary 

This PR migrates the last bit of backend functionality -- updateAllCourseMetrics, which does exactly what it sounds like.

### Test Plan 

Tested manually. Also added new test cases in `AdminActions.test.ts`

### Notes 

It also fixes the odd linter warning in endpoints.ts and a few other places.

### Breaking Changes

Changed some of the return values in AdminActions to the new format they were supposed to be. The endpoints that the frontend used were still the old ones, but we should double check everything works just in case.


AkashAryal (approved):
Looks good. I liked the idea of standardizing the output of the endpoint functions. In the future we should do something similar to do this (perhaps create a general output object/type) for the return value of other endpoints.

PR #239, opened 2021-03-03
Migrate chart function to use axios in frontend

AkashAryal:
### Summary 
Thus PR updates the frontend to use axios to call **getReviewsOverTimeTop15** in place of Meteor. Additionally, the chart is not created unless you click on the button. This should hopefully cause less lag when accessing the admin page.

### Test Plan/Notes

When this function is called, currently i get a proxy error (ECCONRESET) locally. I believe this is due to a timeout. To fix this we might need to update some timeout constants (whether it be on the axios/node/heroku level i'm not sure). That said the migration works as intended. When I change the getReviewsOverTimeTop15 function to simply return the json it would otherwise compute, the chart is displayed in the front end.


khemritolya (commented):
Tested locally. This does seem to massively reduce the slowdown when accessing the admin panel. Could you merge in the changes from master? Looks good otherwise!

khemritolya:
For a view of how drastically this reduced database load, consider the following:

Screenshot_20210307_185016

The first bump is accessing the admin panel with the changes.

The second bump is accessing the admin panel without the changes.

khemritolya (approved):
Looks good

PR #244, opened 2021-03-15
Added Testings

juliejeong:
### Summary 

- Added testing cases that check for queries with no matching results, queries containing non-ascii characters, and queries with subject and number combined (with no space) for **Search.test.ts**
- Added a testing case that calls the method _decrementLikes_ for **Review.test.ts**
- Updated the log message from "getting subjects failed" to "getting professors failed" for the axios call '/v2/getProfessorsByQuery' from **SearchBar.jsx**

### Test Plan 

Newly added tests all passed
Screenshot 2021-03-14 205257 final final


khemritolya (approved):
Looks good! Thanks for improving our coverage!

PR #246, opened 2021-03-20
Remove last frontend meteor calls

khemritolya:
### Summary 

This PR migrates the last bit of Meteor call functionality. In particular, these are some admin actions, some search related functionality, and the reporting functionality. There are all very similar.

It also adds some express endpoints for functions which seem to never have been called. We should look into whether or not we can ditch them.

### Test Plan

Tested locally. Also adjusted backend tests where relevant.

### Notes 

There will need to be another PR to deprecate Meteor subscriptions. There will need to be then the final PR to disable meteor.

There were some Meteor calls to non-existent Meteor methods. These calls have been removed.

### DIFF
--- a/client/src/ui/Admin.jsx
+++ b/client/src/ui/Admin.jsx
@@ -95,53 +95,46 @@ export class Admin extends Component {
   // Call when user asks to remove a review. Accesses the Reviews database
   // and deletes the review with this id.
   removeReview(review, isUnapproved) {
-    Meteor.call('removeReview', review, Session.get("token"), (error, result) => {
-      if (!error && result === 1) {
-        console.log("Review removed");
-        if (isUnapproved) {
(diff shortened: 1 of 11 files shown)


khemritolya:
@AkashAryal It should be good to go. Let me know if you find any issues.

AkashAryal:
When I try reporting a review, I don't see it appear in the admin page. Could you try reporting something  and seeing if it shows up?

khemritolya:
> When I try reporting a review, I don't see it appear in the admin page. Could you try reporting something and seeing if it shows up?

At least for me, it works. Though it took a while to show up. But the db op seemed to have gone through all right.

Screenshot_20210323_124127
Screenshot_20210323_125222
Screenshot_20210323_125311

AkashAryal (approved):

PR #247, opened 2021-03-20
Background time of day changes

tmenezes1:
### Summary 

Added name to developer list in ReadMe.

Made it so that the homepage background changes based on time of day (afternoon, sunset, night).

Homepage background also changes based on season of the year (spring/summer, fall, winter)

### Test Plan 

I ran the code at different times of the day to make sure the background changed.

I changed the month values associated with each season and subsequently ran the code to test for each seasonal background.


client/src/ui/App.tsx
```diff
@@ -2,6 +2,29 @@ import React, { Component } from 'react';
 import SearchBar from './SearchBar.jsx';
 import "./css/App.css";
 
+var date = new Date();
```
khemritolya, line 5: Please to move these lines of code inside a function. Consider the render function later in the file.

client/src/ui/css/App.css
```diff
@@ -35,6 +35,82 @@
   background-position-y: 103%;
 }
 
+.background-gradient_afternoon{
```
khemritolya, line 38: Please factor out the common css style here -- for example background-repeat, background-size, etc.

khemritolya (commented):
Please see the comments -- looks good otherwise!

client/src/ui/App.tsx
```diff
@@ -2,6 +2,29 @@ import React, { Component } from 'react';
...
+var month = date.getMonth();
```
AkashAryal, line 6: use `const` or `let` over `var`

client/src/ui/App.tsx
```diff
@@ -2,6 +2,29 @@ import React, { Component } from 'react';
...
+  monthclass = "winter"
+}
+
+else if (month > 6) {
+  monthclass = "fall"
+}
+
+if (hours < 5 || hours > 19) {
+  dayclass = "night"
+  monthclass = "";
```
AkashAryal, line 21: Is it right to do `monthclass = "";` here when the if statement checks for hours? Perhaps you forgot to delete this?

client/src/ui/css/App.css
```diff
@@ -24,7 +24,7 @@
   margin: 120px 0px 0px 120px;
 }
 
-.background-gradient{
+.background-gradient_summer{
```
AkashAryal, line 27: Does this css get used? Looking at the div we have `background-gradient_" + dayclass + monthclass` and from the looks of it dayclass is never the empty string so it does not seem like this is used?

tmenezes1: It doesn't get used, I'll take it out

client/src/ui/App.tsx
tmenezes1: The reason I set monthclass to the empty string when it is night is that in the figma, there is no design for winter night, only designs for night and winter separately. And combining the night gradient with the winter illustration did not look right because of the bright snow vs the dark sky. So essentially when the time of day is night, it overrides the season illustration and just uses the night illustration, so I set the monthclass to the empty string to only use the night illustration and gradient to avoid the conflicting design.

AkashAryal: Sounds good

juliejeong (commented):
Just some suggestions on the when/how the backgrounds change. Codes look good to me other than the comments already made by others.

khemritolya (approved):
Looks good!

PR #248, opened 2021-03-27
No more subscriptions.

khemritolya:
### Summary 

The time has come to remove Meteor subscriptions. If you want to know why, just take a gander into [`publications.ts`](https://github.com/cornell-dti/course-reviews-react-2.0/blob/master/server/publications.ts). It's pretty bad.

For those things which I thought make sense, we should actually have preserved the live update ability (though I have not tested this all to thoroughly).

### Test Plan 

Tested locally. Verified tests still pass.

### Notes 

I had to add a new backend endpoint, because of the way subscriptions work, the Admin page was fetching the relevant reviews through a subscription, and not an endpoint call. 

### Breaking Changes  

None, everything should still work.

### DIFF
--- a/client/src/index.tsx
+++ b/client/src/index.tsx
@@ -4,10 +4,10 @@ import React from 'react';
 import { Meteor } from "./meteor-shim";
 import { render } from 'react-dom';
 
-import Login              from './ui/Login';
-import ClassView          from './ui/ClassView.jsx';
-import App                from './ui/App';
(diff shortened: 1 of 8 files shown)


AkashAryal (approved):
Maybe add a few tests for the new endpoints. We have tests for all other endpoints so we should probably have some for these as well. Other than that looks good.

AkashAryal (approved):

PR #251, opened 2021-03-29
Modify SearchBar and NavBar for Mobile (new)

juliejeong:
### Summary 

- Changed the navbar so the logo disappears on mobile 
- Changed the size and margins of the search bar to match the mobile design 

### Test Plan 

I ran the program locally with 375 x 812 size

### DIFF
--- a/client/src/ui/css/Navbar.css
+++ b/client/src/ui/css/Navbar.css
@@ -1,9 +1,27 @@
+@media screen and (max-width: 1000px) {
+  .navbar {
+    height: 102px;
+    padding-bottom: 21px;
+    padding-left: 21px;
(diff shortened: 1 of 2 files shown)


tmenezes1 (commented):
Everything looks good. Specifically, the size and formatting of the search bar looks right compared to the mobile design on figma, and the code quality is good. There are only slight differences from the styling on the figma (such as search bar height 42px rather than listed at 46px in the figma) but these are probably insignificant, as it looks identical on the whole.

tmenezes1 (approved):
I tested this locally, everything looks good as far as the spacing and positioning of the search bar and navbar. Code quality looks good as well.

PR #252, opened 2021-03-31
Search results page for mobile 

AkashAryal:
### Summary 

This PR updates the search results page for mobile. More specifically the padding, margins, cards, and border around overall rating for the page were tackled in this PR. The filter button, the top portion of the page (search bar) have not been touched in this PR. Let me know if i missed anything.

image

### Test Plan 

Checked locally by pulling up the page

### Notes 

Some of the figma numbers seemed off (especially the border padding for the overall rating) so i eye-balled it.

### DIFF
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -14,7 +14,10 @@
   </script>
   <script>
     window.dataLayer = window.dataLayer || [];
-    function gtag() { dataLayer.push(arguments); }
+
+    function gtag() {
+      dataLayer.push(arguments);
+    }
     gtag('js', new Date());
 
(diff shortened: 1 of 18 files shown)


khemritolya (commented):
Look good overall. There is the following issue on mobile (i.e. not desktop-pretending-to-be mobile) when using the dropdown. This is on mobile chrome, I can confirm it happens on firefox as well.

a dropdown?

I think this may actually be easier to use than the design, but it is a difference.

khemritolya (approved):
I talked it over with Angela -- this is fine. We may come back to in the future.

PR #253, opened 2021-04-01
Used react-toastify for new notifications

tmenezes1:
### Summary 

Submitting a review now results in either a success or error toast notification, as opposed to alert popups that were used previously.

image
image

### Test Plan 

Tested on successful course review submissions

### DIFF
--- a/client/src/ui/Form.jsx
+++ b/client/src/ui/Form.jsx
@@ -10,6 +10,9 @@ import { Session } from '../meteor-session';
 import { includesProfanity } from "common/profanity";
 import axios from 'axios';
 
+import { toast } from 'react-toastify'
+import 'react-toastify/dist/ReactToastify.css'
+
 /*
   Form Component.
 
@@ -228,6 +231,16 @@ export default class Form extends Component {
     }
   }
 
+  createNotification(text, success) {
+    toast.configure()
+    if (success) {
+      toast.success(text, { position: toast.POSITION.TOP_RIGHT })
+    }
+    else {
+      toast.error(text, { position: toast.POSITION.TOP_RIGHT })
+    }
+  }
+
   submitReview() {
     // Call the API insert function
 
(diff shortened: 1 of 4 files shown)


khemritolya (changes requested):
Looks good overall -- there were some style issues however -- see if you can move it to the center of the screen, and make it larger. In user testing it seems that people miss the popup sometimes if it is in the corner.

juliejeong (commented):
Looks good!! the color is bright so it is easy to catch attention and also the duration of the pop up seems reasonable as well :)

client/src/ui/Form.jsx
```diff
@@ -248,11 +262,14 @@ export default class Form extends Component {
         Session.setPersistent({ "review_num": "" });
         Session.setPersistent({ "courseId": "" });
         this.hide();
-        alert('Thanks for reviewing! New reviews are updated every 24 hours.');
+        //alert('Thanks for reviewing! New reviews are updated every 24 hours.');
```
khemritolya, line 265: I'm not sure why these comments are still hanging around. They appear in a number of other places as well. It is recommended to get rid of unneeded code.

client/src/ui/Form.jsx
```diff
@@ -248,11 +262,14 @@ export default class Form extends Component {
...
+        //toast('Thanks for reviewing! New reviews are updated every 24 hours.')
+        this.createNotification('Thanks for reviewing! New reviews are updated every 24 hours.', true)
       }
       else {
         console.log(res.error);
-        alert("An unknown error occured, please try again.", "danger");
+        //alert("An unknown error occured, please try again.", "danger");
```
khemritolya, line 271: here also

client/src/ui/Form.jsx
```diff
@@ -263,7 +280,8 @@ export default class Form extends Component {
 
   submitError() {
     this.hide();
-    alert("You may only submit a review with a @cornell.edu email address, please try again.", "danger");
+    //alert("You may only submit a review with a @cornell.edu email address, please try again.", "danger");
```
khemritolya, line 283: ditto.

khemritolya (commented):
There are a couple of small issues that I have, see the comments. Seems good otherwise.

khemritolya (approved):
Looks good!

PR #255, opened 2021-04-14
User Review association

khemritolya:
### Summary 

This pull request lays the groundwork for profiles. In particular, it associates users with the reviews they post.

### Test Plan 

Made sure that old tests still pass. Added new tests to verify that we are not leaking user ids.

### Notes 

Even if we leak user ids, there is still no way of turning those back into netIds on the part of the public. Still, better safe than sorry!

### Breaking Changes  

None. It would be bad if there were.


tmenezes1 (approved):
I tested this PR locally: I navigated around the website checking out reviews for specific classes, tried submitting and approving a review, and nothing was broken. I then tested the endpoints for the reviews of a course using curl and could not find a way to see the netids or any user info. Everything looks good.

PR #258, opened 2021-04-21
Mobile Class Review Popup

AkashAryal:
### Summary 
This PR makes the mobile Class Review page (which you get when you click on a card in the search results page). In particular we adapt the whole upper part of the page that includes the gauges. The CourseReviews component (component that actually lists the reviews) has not be adapted in this PR. 

image

If we scroll far enough:
image

### Test Plan 
Checked locally against figma

### Notes 
There are might be a few things that are not quite right. The little "<" on the top left should probably be an icon, but for now I have used the literal "<". If we get an icon from design we can replace it. 

Also I'm not sure if its because I am using a linux machine but some of the fonts seem kind of off (ex: Roboto font on the gauge)


khemritolya (commented):
It does not seem to be working for me on Firefox. In particular, it seems the old review box is still there. Any ideas?

Screenshot_20210421-162753_Firefox

client/src/ui/CourseReviews.tsx
```diff
@@ -26,6 +29,7 @@ export class CourseReviews extends Component<Props, State> {
   constructor(props: Props) {
     super(props);
     this.state = { comparator: "helpful", reviews: [] };
+    //alert(this.props.a)
```
khemritolya, line 32: Is this needed?

client/src/ui/CourseReviews.tsx
```diff
@@ -134,6 +138,9 @@ export class CourseReviews extends Component<Props, State> {
   }
 
   render() {
+
+    //console.log("onscrool", this.props.onScroll);
```
khemritolya, line 142: Ditto

client/src/ui/PreviewCard.jsx
```diff
@@ -136,6 +136,10 @@ export default class PreviewCard extends Component {
   }
 
   render() {
+    //console.log(this.props.mobile);
```
khemritolya, line 139: Ditto

khemritolya (changes requested):
There's one more issue I found: The review thing is cropped on real mobile

a dropdown?

khemritolya (approved):
Looks good now! Nice work!

PR #259, opened 2021-04-21
Mobile review card

tmenezes1:
### Summary 

This pull request formats the mobile review card

### Test Plan 

Tested locally against figma design


khemritolya (changes requested):
As discussed, please to pull in the changes from `mobile-redesign`.

khemritolya (approved):
As discussed via zoom, it all seems to work now.

PR #260, opened 2021-04-23
Added a mobile filter popup

juliejeong:
### Summary 

Added a filter popup for mobile that opens when the filter button is clicked. Changed some CSS files that are used for filters.

### Test Plan 

I ran it locally; it matches the Figma design and all the filtering functionality still works

### Notes 

Note that the "X" button on the upper left of the Figma design is not implemented.


client/src/ui/css/ResultsDisplay.css
```diff
@@ -6,57 +6,194 @@
...
+  }
+}
+
+@media (max-width: 1000px) {
+  .filter-container {
+    margin-top: 36px;
+    margin-left: 24px;
+    margin-right: 24px;
+  }
+}
```
AkashAryal, line 26: You can list multiple css properties inside one `@media` block. So instead of creating new media blocks for every property, create 2 (one for max-width, one for min-width) and put stuff in there.

client/src/ui/css/ResultsDisplay.css
```diff
@@ -6,57 +6,194 @@
...
   font-style: normal;
-  line-height: normal;
+  color: #000000;
   align-items: center;
-  text-decoration-line: underline;
   margin-bottom: 23px;
-  color: #000000;
+}
+
+@media (min-width: 1000px) {
```
AkashAryal, line 35: Another thing here. I noticed that bootstrap ends the medium size at 992 pixels instead of 1000px (the effect of this is clear if you try to set width to between 992 and 1000px - the filter text size gets altered in desktop), so could you change all the 1000px to 992px in this PR? (I think you might have have it that way from the start before we decided to change it haha).

AkashAryal (changes requested):
Looks great! I noted some small code quality changes but its good overall. Good job Julie!

client/src/ui/css/ResultsDisplay.css
juliejeong: I just did! The only problem is that the filter button is there when the width is 991. I'm not sure why exactly this happens.

AkashAryal (approved):
There is some non-trival issue with screen-sizes around some small range. We can tackle that in another PR.

PR #262, opened 2021-05-05
Like/Dislike tracking + frontend bug fix

AkashAryal:
### Summary 
This PR updates schema of `review` and `student` to allow for tracking of likes/dislikes and updates the `incrementLike` and `decrementLike` functions to handle the changes.

A small bug dealing with an onScroll function being undefined in desktop was also fixed.

### Test Plan 
Tested locally by liking/disliking.

### DIFF
--- a/server/endpoints/Review.ts
+++ b/server/endpoints/Review.ts
@@ -271,7 +271,8 @@ export const decrementLike: Endpoint<ReviewRequest> = {
         await Reviews.updateOne({ _id: request.id }, { $set: { likes: 0, lastDislikedIP: ctx.ip } }, { $pull: { likedBy: student.netId } }).exec();
       } else {
         // bound the rating at 0
-        await Reviews.updateOne({ _id: request.id }, { $set: { likes: Math.max(0, review.likes - 1), lastDislikedIP: ctx.ip } }, { $pull: { likedBy: student.netId } }).exec();
+        await Reviews.updateOne({ _id: request.id },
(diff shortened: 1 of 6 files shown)


server/endpoints/Review.ts
```diff
@@ -196,22 +197,37 @@ export const insertReview: Endpoint<InsertReviewRequest> = {
...
+
+        if (student.likedReviews !== undefined && student.likedReviews.includes(review.id)) {
+          return { resCode: 0, error: "Error: user already liked" };
+        }
+
+        // if likedReviews is undefined its fine. $push creates empty arr
+        await Students.updateOne({ netId: student.netId }, { $push: { likedReviews: review.id } });
+
+        if (review.likes === undefined) {
+          await Reviews.updateOne({ _id: request.id }, { $set: { likes: 1, lastLikedIP: ctx.ip }, $push: { likedBy: student.id } }).exec();
```
khemritolya, line 223: I think we can stop tracking IPs now.

khemritolya (changes requested):
Mostly good save for the comment. There's also the issue of if you have not gotten the sign on token, it will not let you like anything. It should probably prompt you, like how it currently does for when you leave a review.

server/endpoints/Review.ts
khemritolya: Occurs in some other places as well

khemritolya (approved):
Looks good!

PR #269, opened 2021-05-15
Remove cypress and dumprestore.sh

AkashAryal:
### Summary 

This pull request removes `cypress` and `dumprestore.sh`. 

In particular we remove `cypress` by updating `.gitignore`, `cypress/*` , `package.json`

### Test Plan 
Ran locally, nothing seems broken.


khemritolya (approved):
Sad that cypress did not work out, but it is what it is. Looks good!

PR #271, opened 2021-05-15
removed major search

juliejeong:
### Summary 

This PR removes the major search function

### Test Plan 

Testing locally and the filter function still works as intended


khemritolya (approved):
Looks good!

PR #272, opened 2021-05-15
Updated spring backgrounds

juliejeong:
### Summary 

This PR updated the spring backgrounds because there was a minor design issue with the original images.

### Test Plan 

I tested this locally and all three updated spring backgrounds are displayed properly at appropriate times.


AkashAryal (approved):
Good catch!

PR #273, opened 2021-05-15
Fixes screen sizing and gage sizing

tmenezes1:
### Summary 

This pull request fixes problematic screen width sizes that were in the range of 990px - 1000px. It also manages the height of the gauges so that it does not overflow and intersect with anything. Note: width 992px is problematic and does not work.

### Test Plan 

Tested locally on multiple screen sizes with widths in the range of 990 to 1000px


khemritolya (changes requested):
Please to merge in the changes from the main branch,

khemritolya (commented):
Currently attempting to debug this behavior, which only seems to occur on firefox. (Notice the gauge bar is smaller than on production).
Screenshot_20210517_161000

kungpaogao:
Fixed in #279

PR #274, opened 2021-05-15
Small padding fix to dropdown menu

AkashAryal:
### Summary 
This pull request adds 8px bottom padding to dropdown menu

### Test Plan 
Ran locally


juliejeong (approved):
Tested locally. Now there is a margin and it seems nice! :)

PR #282, opened 2021-10-07
Upgrading to Bootstrap 4

jack-farley:
### Summary 

This pull request updates CU Reviews from bootstrap 3 to bootstrap 4. Currently, the client is largely updated, except for the class view page, which Andrew is updating and will add on top of this PR. 

- [x] Upgrades to bootstrap 4
- [x] Removes panels
- [x] Updates everything to use the new version of the grid system
- [x] some other minor changes

### Test Plan 

I have tested these updates on many screen sizes from a large monitor to a small window, but I haven't done too much testing on mobile yet.

### Linter Warnings 

Right now I am not seeing any linter warnings for the update. 

### Notes

If you are getting an error after pulling this branch, try removing your node modules and reinstalling.


client/src/ui/ClassView.jsx
```diff
@@ -118,7 +118,7 @@ export class ClassView extends Component {
       );
       return (
         <div>
-          <div className="hidden-lg hidden-xl hidden-md">
+          <div className=".d-none.d-xs-block.d-md-none">
```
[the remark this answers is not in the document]
jack-farley: Thanks, good catch. Wasn't working when I tested it, wasn't sure why - this seems to have fixed it though.

client/src/ui/PreviewCard.jsx
```diff
@@ -186,13 +186,13 @@ export default class PreviewCard extends Component {
 
         {!this.props.transformGauges && (
           <div className="row gaugeHolder">
-            <div className="col-md-4 col-sm-4 col-xs-4 remove-left-padding">
+            <div className="col-lg-4 col-md-4 col-sm-4 col-xs-4 remove-left-padding">
```
kungpaogao, line 189:
```suggestion
            <div className="col-lg-4 col-md-4 col-sm-4 col-4 remove-left-padding">
```
there's no more `xs` breakpoint, so here and elsewhere `col-xs-{value}` should just be `col-{value}`

client/src/ui/UpdateReview.tsx
```diff
@@ -90,33 +90,39 @@ export default class UpdateReview extends Component<Props, State> {
...
-              <div className="panel panel-default">
-                <div className="panel-body text-center">{review.difficulty}</div>
+              <div className="col-md-3 col-sm-6 col-xs-6">
+                <div className="card-body"> Overall Rating</div>
+                <div className="card-body"> Difficulty</div>
+                <div className="card-body"> Professor(s)</div>
               </div>
-              <div className="panel panel-default">
-                <div className="panel-body text-center">{review.professors}</div>
+              <div className="col-md-6 col-sm-12 col-xs-12">
```
kungpaogao, line 118: `col-xs` --> `col`

kungpaogao (changes requested):
overall, looks great - looks like this was pretty tedious to do!

biggest thing I noticed (without checking out the branch locally) is that `col-xs` should be just `col` (https://getbootstrap.com/docs/4.0/migration/#grid-system)

AkashAryal (approved):
Looks Good! Thanks for the big PR!

kungpaogao (approved):
looks good :))

PR #280, opened 2021-10-06
Counting user reviews

michelleli01:
### Summary 
I created a new Profile.ts file to store all the backend endpoints for collecting user profile data and implemented the endpoint for counting all reviews made by a particular user. The countReviewsByStudentId endpoint queries through the MongoDB database to find all the reviews posted by a particular user with a certain studentId. Then, countReviewsByStudentId counts the number of reviews made by said student. 

### Test Plan 
I used Postman to send requests to the database and ensure that the number returned is accurate. I created an automated test suite to run tests.

PR #291, opened 2021-10-27
Counting total number of likes on a student's reviews

juliejeong:
### Summary 
I created a new endpoint _getTotalLikesByStudentId_ that takes a student's net id and returns the total number of likes the student has gotten on their reviews.

### Test Plan 

I created a test file with testing data of students and reviews and checked that the endpoint correctly returns the total number of likes.

### DIFF
--- a/server/endpoints/Profile.test.ts
+++ b/server/endpoints/Profile.test.ts
@@ -118,19 +118,6 @@ const testUsers: Student[] = [
     reviews: null,
     likedReviews: [],
   },
-  /*
-  {
-    _id: "testid",
-    firstName: "John",
-    lastName: "Smith",
-    netId: "js0",
-    affiliation: null,
-    token: null,
-    privilege: "regular",
-    reviews: ["4Y8k7DnX3PLNdwRPr", "4Y8k7DnX3PLNdwRPq", "3yMwTbiyd4MZLPQJF"],
-    likedReviews: [],
-  },
-  */
 ];
 
 beforeAll(async () => {
--- a/server/endpoints/Profile.test.ts
+++ b/server/endpoints/Profile.test.ts
@@ -1,0 +1,117 @@
+/* eslint-disable import/prefer-default-export */
+import mongoose from 'mongoose';
+import { MongoMemoryServer } from 'mongodb-memory-server';
+import express from "express";
+
+import axios from 'axios';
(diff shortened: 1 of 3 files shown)


server/endpoints/Profile.ts
```diff
@@ -0,0 +1,36 @@
...
+      const reviews: ReviewDocument[] = await Promise.all(
+        (reviewIds).map(async (reviewId) => await Reviews.findOne({ _id: reviewId })),
+      );
+      reviews.forEach((review) => {
+        if ('likes' in review) {
+          totalLikes += review.likes;
+        }
+      });
+
+      return { resCode: 1, totalLikes };
```
AkashAryal, line 27: Lets start standardizing the return object of our end points. In Michelle's PR that was merged in earlier (https://github.com/cornell-dti/course-reviews-react-2.0/pull/280/files) we went with `{code: <> , message:<>}` where message contains data if there is no error. Lets try that here too

server/endpoints/Profile.ts
```diff
@@ -0,0 +1,36 @@
...
+    } catch (error) {
+      // eslint-disable-next-line no-console
+      console.log("Error: at 'getTotalLikesByStudentId' method");
+      // eslint-disable-next-line no-console
+      console.log(error);
+      return { resCode: 0 };
```
AkashAryal, line 33: Same here

AkashAryal (changes requested):
Looks good overall Julie! Just a few small consistency changes. Good job!

server/endpoints/Profile.ts
juliejeong: fixed!

server/endpoints/Profile.test.ts
```diff
@@ -68,8 +118,27 @@ const testUsers: Student[] = [
     reviews: null,
     likedReviews: [],
   },
+  /*
```
AkashAryal, line 121: Let's delete if we don't need this

AkashAryal (changes requested):
Looks good! Just delete the commented out code and it should be good. The tests pass locally but we seem to get a mongo index error here in ci. I will probably just override and force merge it.

server/endpoints/Profile.test.ts
juliejeong: Fixed!

PR #292, opened 2021-10-29
Login Popup

jack-farley:
### Summary 

This pull request creates the login popup that allows users to login on the website without having to leave a review. Currently, there is no way to tell that you are logged in, except that you will not have to log back in to leave a review. I think a good next step would be to replace the sign in button in the top corner once a user is signed in.

- [x] login button
- [x] login popup
- [x] implement backend functionality

### Test Plan 

Capture

Capture-popup

### Notes 

This pull request is working off of the bootstrap 4 upgrade.


michelleli01 (commented):
Great job with getting the login pop up to work by using modals! I ran it locally on my computer and functionality worked very smoothly and your code was very easy to read :)

PR #293, opened 2021-11-01
left profile card 

michelleli01:
### Summary 
I implemented the Figma design for the left profile card part of the user profile page. Right now, the data and information are hard coded in and I have not implemented/connected the backend portion.

- [x] user statistics
- [x] user profile information
- [x] sign out button
- [x] styled components
- [ ] implement backend functionality

### Test Plan 
image


client/src/ui/css/ProfileCard.module.css
```diff
@@ -0,0 +1,54 @@
...
+  margin-left: 5%;
+
+  color: #000000;
+}
+
+.profileCardImage {
+  margin-left: 10%;
+}
+
+@import url("https://fonts.googleapis.com/css2?family=Roboto&display=swap");
```
kungpaogao, line 54: put `@import` at top of file

kungpaogao (changes requested):
Great use of CSS Modules, functional components, and TypeScript :)

It seems like a lot of styles are from Figma, so I would make sure to test that these behave as expected on different screen sizes. I think you might run into some issues because some width/heights are static values.

kungpaogao (approved):
looks good!!

AkashAryal (changes requested):
There is a conflict in a few files. After you fix that we should be good!

PR #287, opened 2021-10-20
Update ClassView layout and styles (1/3 for rewriting ClassView)

kungpaogao:
### Summary 

This pull request is the first step towards rewriting the `ClassView` page. This PR deals with the layout and styling of `ClassView`.

- created new component called `ClassViewNew` to house changes for now
- removed lots of nesting of components and styles (e.g., `ClassView` directly imports and uses `Form` component)
  - this allows for better layout management - breakpoints and scaling can be set in parent CSS
- redid `Gauge` layout and styling again to make alignment easier
- fixed lots of small visual glitches (e.g., alignment of scroll bar on right side of course reviews, gauge container right-side alignment)
- updated overflow behavior to be more predictable
- removed `CourseCard` component because it's useless
- see **Breaking Changes** section below for more
- more context in Notion: https://www.notion.so/Rewrite-ClassView-component-76aabbe867194c2387f592279f98017e

### Test Plan 

Tested on macOS: Firefox Developer Edition, Safari, Chrome. Tested on iOS: Safari.

macOS
| old | new |
| - | - |
| image | image |

iOS
| old | new |
| - | - |
| image | image |
- top bar doesn't collapse when scrolling in new implementation (see **Breaking Changes**)
- missing scrollbar
- missing "Leave a Review" button

### Linter Warnings 

n/a

### Notes 

- DO NOT MERGE this into `master` branch - this fix is foundational work for following PRs that will be merged into `kungpaogao/class-view-rewrite`, which will then be merged into `master` as a package
- all this work will probably be rebased onto @jack-farley's Bootstrap 4 work
- we should rethink spacing and how to handle it consistently - right now, I am using pretty arbitrary numbers like `70px` and `35px` based on previous work, but this should be standardized
  - maybe we should look into using CSS variables?? 

### Breaking Changes  

- this current PR is missing several features
  - missing ability to add review on mobile - no buttons or form
  - removed `ResultsDisplayMobile` (for now)
  - missing scroll interactions (e.g., collapsing top bar on mobile)
  - missing past review count and sorting feature for reviews
- some bugs
  - fixed nested `.container` and `.row` usage in `index.tsx` and children components - this causes the `ResultsDisplay` header to shift left because of negative margins
  - `Form` component doesn't scale super well at the moment, but that will be addressed in future PRs (probably 3/3)

PR #294, opened 2021-11-03
Move fetching logic from CourseReviews to ClassView and stop using ResultsDisplayMobile (2/3 for rewriting ClassView)

kungpaogao:
### Summary 

This pull request is another step towards rewriting `ClassView` and making code more readable. 

- remove API requests from `CourseReviews` - makes it really simple
- move fetching, sorting, and reporting review logic into `ClassView`
- moved styles to `ClassView.module.css`
- basically deprecate `ResultsDisplayMobile`

Task link: https://www.notion.so/Move-fetching-logic-from-children-to-ClassView-f2455c306edb4423b80746d0924790fc. See #287 for more context.

### Test Plan 

- tested all the functions (fetching, sorting, reporting) in new implementation
- made sure that you can click through all results on mobile

### Linter Warnings 

n/a

### Notes 

- generally, this should be a model for future pages/components
- it might be worth it to reformat our file/folder structure to encourage this with `pages/` and `components/`
- also, future work should abstract API requests to a separate file like `api.ts`


client/src/ui/CourseReviews.tsx
```diff
@@ -1,199 +1,30 @@
...
-          this.setState({ reviews: this.renderReviews(reviews) });
-          this.sort(this.state.comparator);
-        } else {
-          // eslint-disable-next-line no-console
-          console.log(`Unable to find reviews by course by id = ${this.props.courseId}`);
-        }
-      });
-    }
-  }
-
```
jack-farley, line 61: I like moving the fetching and sorting logic out of CourseReviews - seems to make the component much cleaner.

client/src/ui/CourseReviews.tsx
```diff
@@ -1,199 +1,30 @@
...
 
-  // Handles selecting different sort bys
-  handleSelect = (event: any) => {
-    let opt = event.target.value;
-    this.setState({ comparator: opt });
-    this.sort(opt);
+export default function CourseReviews({
+  reviews,
```
jack-farley, line 11: You've probably already seen this, but it seems like ResultsDisplayMobile also uses the CourseReviews component and is currently failing to build since it does not have the logic to pass in a list of reviews. You mentioned in the PR that it might be a good idea to store api calls in a separate file (api.ts) - seems like moving the fetching and sorting logic from ClassView to a file like that could be a possible solution here.

jack-farley (changes requested):
This looks awesome Andrew! The refactor makes the CourseReviews and ClassView components much cleaner. Once ResultsDisplayMobile is updated to account for this change, I think it should be good to go.

client/src/ui/CourseReviews.tsx
kungpaogao: good catch - I fixed this now, so hopefully it builds

AkashAryal (approved):
I was able to get it to build locally now! CourseReviews looks much nicer now. Gj!

jack-farley (approved):
Looks great, everything seems to work now!

PR #295, opened 2021-11-10
Fixing Search By Professor

jack-farley:
### Summary 

This pull request fixes search by professor by removing the '+' character from queries and switching to use a case insensitive regex.

### Test Plan 

Just did quick user testing for this.


AkashAryal (approved):
Thanks for the fix. This was a widely requested fix :)

PR #301, opened 2021-11-22
Get Reviews by Student

juliejeong:
### Summary 

This PR added an endpoint [getReviewsByStudentId] in Profile.ts that returns the list of Review objects for the given student's netID. 

### Test Plan 

Tested locally. Since there is a discrepancy between the test review objects and review objects generated by the database, the number of returned reviews is compared for convenience.

### DIFF
--- a/server/endpoints/Profile.test.ts
+++ b/server/endpoints/Profile.test.ts
@@ -1,112 +1,166 @@
+
 /* eslint-disable import/prefer-default-export */
 import mongoose from 'mongoose';
-import { MongoMemoryServer } from 'mongodb-memory-server';
-import express from "express";
 
 import axios from 'axios';
 import { TokenPayload } from 'google-auth-library';
 
 import { Review, Student, Class, Subject, Professor } from 'common';
-import { configure } from "../endpoints";
-import { Classes, Reviews, Students } from "../dbDefs";
 import * as Auth from "./Auth";
 
 import TestingServer, { testingPort } from './TestServer';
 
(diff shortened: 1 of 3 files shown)


server/endpoints/Profile.test.ts
```diff
@@ -140,7 +176,7 @@ const mockVerificationTicket = jest.spyOn(Auth, 'getVerificationTicket')
   .mockImplementation(async (token?: string) => validTokenPayload);
 
 afterAll(async () => {
-  await mockVerificationTicket.mockRestore();
+  mockVerificationTicket.mockRestore();
```
kungpaogao, line 179: why did we remove `await` here?

AkashAryal: It looks like the method is returns void so we didnt need to have await there

AkashAryal (changes requested):
Looks good! A small thing so that we don't have repeated code: could you create a function that returns the review objects when given a netid and use that here and in the other endpoint function? That way we don't have duplicate code

juliejeong:
> Looks good! A small thing so that we don't have repeated code: could you create a function that returns the review objects when given a netid and use that here and in the other endpoint function? That way we don't have duplicate code

changed!

AkashAryal (approved):
Looks good!! Thanks for the quick update Julie <3

michelleli01 (commented):
The code is easy to read and the endpoint does get all the reviews by a particular student based on netId. tThe test cases are thorough, well written, and all pass when I run it locally.

PR #304, opened 2021-12-05
Remove profanity filter

AkashAryal:
### Summary 

This PR removes the profanity filter. Also removed mongolab from app.json since we dont use that anymore.

Reasoning:
The profanity package we use is too strict and in general most packages will probably be this way, This causes [bad user experience](https://www.reddit.com/r/Cornell/comments/r8z2hy/cureviews_website_for_cornell_course_reviews/). Additionally, we already manually check reviews so this filter is unnecessary in most cases. 

### Test Plan 

Tested manually.


kungpaogao:
lol i saw this reddit thread too -- looks good!! hopefully deployments are fixed too :))

kungpaogao (approved):

PR #310, opened 2021-12-18
404 Not Found Page

juliejeong:
### Summary 

This PR creates a 404 not found page in `NotFound.tsx` with CSS properties defined in `NotFound.modules.css`. BrowserRouter is also modified in `index.tsx` to use Switch. All the URLs that are not matched to the specified ones will be redirected to this page. **mobile version is not implemented yet

image

### Test Plan 

Tested locally: when an unrecognized URL is entered, the above page shows up while everything else's functionality is unchanged.


client/src/ui/css/ResultsDisplay.css
```diff
@@ -420,3 +420,15 @@ ul {
...
+ 
+.not-found-padding-top{
+  padding-top: 90px;
+}
+.not-found-text {
+  font-family: "Source Sans Pro", sans-serif;
+  font-size: 50px;
+  font-weight: var(--regular-weight);
+  text-align: center;
+}
```
kungpaogao, line 434: we should use [CSS Modules](https://github.com/css-modules/css-modules):
1. create separate file under: `css/NotFound.module.css`
2. move styles there (might have to rename)
3. import in `NotFound` component file
```jsx
import styles from "./css/NotFound.module.css"
```
4. use it by writing (for example) `<div className={styles.notFoundClassName}> ...` 

You can see a good example of this in Michelle's PR: https://github.com/cornell-dti/course-reviews-react-2.0/pull/293/files

client/src/ui/NotFound.tsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+
+export default function NotFound() {
+  return (
+    <div className="col-md-12 col-sm-12 col-xs-12 not-found-padding-top">
```
kungpaogao, line 5:
```suggestion
    <div className="col-12 not-found-padding-top">
```

since we upgraded to Bootstrap 4 this semester, `col-xs-12` is now just `col-12` (see the [migration guide](https://getbootstrap.com/docs/4.0/migration/#grid-system) and the [Bootstrap 4 grid options documentation](https://getbootstrap.com/docs/4.0/layout/grid/#grid-options)). It also looks like you're applying the full column width to all screen sizes, so you can simplify it to just `col-12`.

client/src/ui/NotFound.tsx
```diff
@@ -0,0 +1,11 @@
...
+      <img src="/noResults.svg" className="img-responsive no-results" alt="No class found" ></img>
```
kungpaogao, line 6:
```suggestion
      <img src="/noResults.svg" className="img-fluid no-results" alt="No class found" ></img>
```
see [Bootstrap 4 migration guide](https://getbootstrap.com/docs/4.0/migration/#images)

kungpaogao (changes requested):
good job overall! 404 pages are pretty important for a good user experience

I requested some changes related to Bootstrap 4 upgrades and CSS Modules

client/src/ui/css/ResultsDisplay.css
juliejeong: Thank you for your feedbacks! I made changes.

client/src/ui/NotFound.tsx
juliejeong: Fixed! Thanks

client/src/ui/NotFound.tsx
juliejeong: Fixed!

client/src/ui/css/NotFound.module.css
```diff
@@ -0,0 +1,25 @@
...
+  height: 300px;
+  margin-left: auto;
+  margin-right: auto;
+  margin-bottom: 10px;
+}
+
+.notFound404Text {
+  font-family: "Source Sans Pro", sans-serif;
+  font-size: 50px;
+  font-weight: var(--regular-weight);
```
kungpaogao, line 16: great job converting this its own CSS Module! The annoying part containerizing CSS code is that you can't use global variables without importing files.

so, you need to add this at the top of the file (you probably won't see a visual change in the UI because we are just setting `font-weight` to a default value): 
```css
@import "../../index.css";
```

kungpaogao (commented):
great job fixing this so quickly! I added some other comments on the code, but I have yet to try it out locally—I'll approve it after I finish that

jack-farley (approved):
Great job! The not found page looks great, should be a big improvement over the blank page we have currently.

michelleli01 (commented):
Great job on the not found page. It's definitely an improvement in user experience. Your use of CSS modules makes your code incredibly easy to read. Keep up the great work!

AkashAryal (approved):
Thanks for implementing this! Looks great

kungpaogao (approved):
Great job, looks good!

PR #320, opened 2022-02-24
Review Rewrite

jack-farley:
### Summary 

This pull request rewrites the Review component. It also adds support for pending reviews and reviews on the profile page (which need to include course titles).

- [x] rewrote Review.jsx in ReviewNew.tsx
- [x] added styling for pending reviews
- [x] added styling for the profile page

### Test Plan 

I have user tested this pull request on several screen sizes. Pictures are below:

Screen Shot 2022-03-03 at 6 39 27 PM
Screen Shot 2022-03-03 at 6 40 14 PM
Screen Shot 2022-03-03 at 6 41 36 PM
Screen Shot 2022-03-03 at 6 42 11 PM
Screen Shot 2022-03-03 at 6 43 12 PM
Screen Shot 2022-03-03 at 6 46 35 PM

### Linter Warnings 

There are currently no linter warnings.

### Notes 

- The like button is still broken - problem with that seems to be that expired tokens are not refreshed on the client side, but I think that should be fixed in a different PR.

- The pending review styling is used automatically when a review is passed in with the 'visible' field set to false - this may be a problem on the admin page.


michelleli01 (commented):
The new review formatting looks great! The code is very readable and it displays on my laptop perfectly fine. The way the title, professor, and review align exactly with the ratings section on mobile devices looks kinda strange to me. So, it would be nice to add some padding within the `contentContainer` class so there is a distinction between the two sections on small screens. Great work!

client/src/ui/CourseReviews.tsx
```diff
@@ -19,10 +19,11 @@ export default function CourseReviews({
     <div>
       {reviews.map((review) => (
         <ReviewCard
-          key={review._id}
-          info={review}
+          reviewId={review._id}
```
kungpaogao, line 22: I would be careful removing the `key` here because it's used in the `map()` to tell React what item to update (see here: https://beta.reactjs.org/learn/rendering-lists#why-does-react-need-keys)

client/src/ui/CourseReviews.tsx
```diff
@@ -19,10 +19,11 @@ export default function CourseReviews({
...
+          review={review}
           reportHandler={reportReview}
           isPreview={false}
+          profilePage={false}
```
juliejeong, line 26: since `isPreview ` is the name we are using for another field, maybe we can name this something similar like `isProfile` for coherence?

juliejeong:
The new review cards' designs have been implemented well! I really love how you managed to have different styles applied to reviews depending on their status rather than creating two CSS styles. This makes everything so much easier. I can't wait to see this being merged and actually applied to cards!!

PR #326, opened 2022-03-14
Fix relevance of search bar suggestions

kungpaogao:
### Summary 

The search bar shows somewhat irrelevant search suggestions. For example, searching in our current production website exhibits two main issues:

1. the suggestions aren't sorted correctly and don't match with the search results on the `/results` page
2. searching the same term can lead to different suggestions depending on typing and network speed

Example of issue 1: Searching "operating systems" shows pretty irrelevant suggestions in the search bar, but the actual results returned in the `/results` page are much more relevant.
image

Example of issue 2: Typing in "2110" multiple times into the search bar leads to different suggestions each time.
| 1 | 2 | 3 |
| --- | --- | --- |
| image | image | image |

This PR fixes both these issues by:

1. sorting suggestions based on their relevance score and then by course number (this is what the `/results` page components do)
2. debouncing search input and cancelling network requests when user inputs a new search string

### Test Plan 

Here are some sample searches before and after the changes.
| before | after |
| --- | --- |
| image | image |
| image | image |

### Linter Warnings 

None.

### Notes 

I have set debounce time to `200ms` because this seems to be a good balance between not updating suggestions and minimizing async issues. See some references here:

- https://ux.stackexchange.com/questions/95336/how-long-should-the-debounce-timeout-be
- https://blog.logrocket.com/how-and-when-to-debounce-or-throttle-in-react/

Right now, we render majors and professors above classes regardless of their similarity scores. Should we consider reordering these according to their similarity score?
image

Additionally, should we consider sorting on server?

Read more notes on the Notion doc here: https://www.notion.so/cornelldti/Fix-search-features-f9a29d71c9dc436baf78a2fbbf226f26

### DIFF
--- a/client/src/ui/Course.tsx
+++ b/client/src/ui/Course.tsx
@@ -37,9 +37,13 @@ export default class Course extends Component<Props> {
       }
     // check if a query was provided, if so underline parts of the class name
     if (this.props.query) {
-      if (text.toLowerCase().indexOf(this.props.query) !== -1) {
-        const startIndex = text.toLowerCase().indexOf(this.props.query);
-        const endIndex = startIndex + this.props.query.length;
+      const textLowerCase = text.toLowerCase();
+      const queryLowerCase = this.props.query.toLowerCase();
+      const queryIndex = textLowerCase.indexOf(queryLowerCase)
+
+      if (queryIndex !== -1) {
+        const startIndex = queryIndex;
+        const endIndex = startIndex + queryLowerCase.length;
         text = <span className="ellipsis">{text.substring(0,startIndex)}<span className='matching-text'>{text.substring(startIndex,endIndex)}</span>{text.substring(endIndex)}</span>
       } else {
         // based on search technique in server/publications, results without a contains match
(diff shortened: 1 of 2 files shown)


juliejeong:
This is such a nice fix that definitely enhances the user experience. I love it!! For the major and professor in the search results, I think those are not always the priority and we can reorder them. Or maybe we could modify our search bar so that users can input what category they are specifically searching for.

juliejeong (commented):
This is such a nice fix that definitely enhances the user experience. I love it!! For the major and professor in the search results, I think those are not always the priority and we can reorder them. Or maybe we could modify our search bar so that users can input what category they are specifically searching for.

AkashAryal:
One thing i notice is that when searching for something like "machine learning" the third result has "6888: Deep" highlighted. In this case we should probably have either nothing highlighted or just the "learning" portion.

kungpaogao:
@AkashAryal I see that when searching something like "operating systems" for courses with "systems" in their name, but I don't see it for "machine learning" after fixing the sorting

image

| this branch | production |
| --- | --- |
| image | image |

Coreball:
This is a great fix, love it

AkashAryal (approved):
This is great! Our searching works a lot better now. Thanks for the hard work Andrew!

PR #327, opened 2022-03-21
[Old, Don't Merge] Fixing the Update Professors Button

jack-farley:
### Summary 

This pull request fixes the update professors button on the admin console. The problem is that the frontend is trying to use expired tokens to authenticate with the backend, so this PR removes expired tokens from its session.

### Notes 

This problem is also causing the like button on reviews not to work, so this fix should fix that as well.


AnkitLakkapragada (commented):
I like how simplified the code is by creating getToken(). I was a little confused about the reason why we were changing resCode 1 from 0 in AdminActions.ts. Looks good otherwise and I look forward to seeing this finished!

AkashAryal:
@jack-farley Should this be closed?

PR #329, opened 2022-03-21
[WIP] Sending Emails Set Up

juliejeong:
### Summary 
[WORK IN PROGRESS] This PR sets up the framework that allows the authorized admin to send emails to all the registered users. This feature will be used towards the end of the semester.

### Test Plan 
A test email will be sent to a temporary email list and will be checked for any potential bugs. 

### Breaking Changes  
Will be added if applicable


AnkitLakkapragada (commented):
So far it looks good! I just had a small question - have you considered the possibility that the service is not gmail and some other provider and what would happen then?

PR #338, opened 2022-04-18
Fixing Update Professors

jack-farley:
### Summary 

This pull request fixes the update professors button on the admin page by discarding expired tokens and refetching them.

### Test Plan 

I tested this with user testing.

### Linter Warnings 

No linter warnings.

### Notes 

- This PR fixes authentication on the admin page, but does not address it on the main site. It should probably be followed by another PR to fix authentication across the site.

### DIFF
--- a/client/src/auth_utils.ts
+++ b/client/src/auth_utils.ts
@@ -12,7 +12,7 @@ export const setAuthToken = (token: string) => {
 export const getAuthToken = () => {
     const token = Session.get("token");
     const exp = JSON.parse(atob(token.split(".")[1])).exp;
-    if (token && token !== "" && exp < Math.floor(Date.now() / 1000)) {
+    if (token && token !== "" && exp > Math.floor(Date.now() / 1000)) {
         return token;
     }
     else return null;
--- a/client/src/auth_utils.ts
+++ b/client/src/auth_utils.ts
@@ -1,0 +1,19 @@
+import { Session } from './session-store';
+
+export const setAuthToken = (token: string) => {
+    Session.setPersistent({ "token": token });
+    if (Session.get("token") !== token) {
+        console.log("Error saving token to session")
+        return false;
+    }
+    return true;
+}
+
+export const getAuthToken = () => {
+    const token = Session.get("token");
+    const exp = JSON.parse(atob(token.split(".")[1])).exp;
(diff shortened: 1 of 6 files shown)


AnkitLakkapragada (commented):
I liked the way you fixed this bug !! However, I was wondering if you still needed the `        console.log(response.data);` in Admin.jsx or if that was simply for debugging purposes.

server/endpoints/AdminActions.ts
```diff
@@ -138,7 +138,7 @@ export const setProfessors: Endpoint<AdminProfessorsRequest> = {
       const userIsAdmin = await verifyToken(adminProfessorsRequest.token);
       if (userIsAdmin) {
         const semesters = findAllSemesters();
```
AkashAryal, line 140: add await here too

AkashAryal (approved):
Thanks for taking care of the auth issue here for admin page!! 

This is a good first step towards fixing update professors. I was not able to actually update professors yet though. There is another error after you add the await. 

There should probably be two more PRs after this one. 
1. One PR for fixing the issue i mentioned 
2. Another PR to update auth in non admin pages as you mentioned.

1 takes a bit of priority over 2 since we want to be able to update professors as soon as possible. 

A few things i discovered while testing this PR that might be helpful. 
* To test if fixing professors works properly, you can check if "Eshan" is a professor option for cs 4820. He currently is not listed. 
* Currently updateProfessors loops over all semsters from FA15 to now. This takes a while. For testing purposes i recommend hard coding `const semesters = ['FA22'];` in setProfessors endpoint in AdminActions.ts instead of calling findAllSemesters().

client/src/auth_utils.ts
```diff
@@ -0,0 +1,19 @@
...
+        console.log("Error saving token to session")
+        return false;
+    }
+    return true;
+}
+
+export const getAuthToken = () => {
+    const token = Session.get("token");
+    const exp = JSON.parse(atob(token.split(".")[1])).exp;
+    if (token && token !== "" && exp > Math.floor(Date.now() / 1000)) {
```
AkashAryal, line 15: Is this right? I think when Michelle tried using this for profile menu it was not fully working.

AkashAryal (approved):
Thanks for the fix Jack. This is super important so that we have up to date professors.

PR #340, opened 2022-04-18
Implementing Grade & Major Options in ReviewForm and Review

AnkitLakkapragada:
### Summary 

This pull request is implementing Grade received and Major options to the Review form and Reviews. These are optional fields that users can choose to submit in their review. This was implemented into the newly written ReviewForm.tsx. 

### Linter Warnings 
None

### Notes 

This feature makes the review form longer. This means that a user has to scroll all the way down to submit a review. To fix this, future steps involve a pop-up form or possibly bringing back double-column scrolling.


client/src/ui/Form.jsx
```diff
@@ -326,6 +351,96 @@ export default class Form extends Component {
...
+    this.gradeReceived = ["A+", "A", "A-", "B+", "B", "B-", "C+", "C", "C-", "D+", "D", "D-", "F"];
+    return this.gradeReceived.map((grade) =>
+      ({
+        "value": grade,
+        "label": grade
+      }))
+  }
+
+getMajorOptions() {
+    this.majorsOptions = ["Computer science",
```
AkashAryal, line 364: Lets put these in another file and import from there. I see that this list is repeated later on in code too. We dont want that if possible.

server/endpoints/Search.ts
```diff
@@ -223,7 +223,8 @@ export const getCoursesByProfessor: Endpoint<Search> = {
       let courses = [];
       const regex = new RegExp(/^(?=.*[A-Z0-9])/i);
       if (regex.test(search.query)) {
-        courses = await Classes.find({ classProfessors: search.query }).exec();
+        const professorRegex = search.query.replace('+', '.*.');
+        courses = await Classes.find({ classProfessors: { $regex: professorRegex, $options: "i" } }).exec();
```
AkashAryal, line 227: what is this for?

AkashAryal (changes requested):
Looks good over all Ankit! This feature will be useful and make the reviews more impactful for users. Left a few comments.

server/endpoints/Search.ts
AnkitLakkapragada: I don't know, I did not implement this. Looks to be for finding the professors for a course?

client/src/ui/Form.jsx
AnkitLakkapragada: Should I delete Form.jsx where this is duplicated?

server/endpoints/Search.ts
kungpaogao: might have to rebase

client/src/ui/Form.jsx
kungpaogao: I think it might be a good idea to delete `Form.jsx` and other unused files in another PR

AkashAryal (approved):
Thanks for the PR Ankit, looks good! Adding major and grade received should make reviews have more impact. 

Edit: It seems like some grades like A+ show up as N/A

PR #341, opened 2022-04-18
Enable user to add unlisted professors

kungpaogao:
### Summary 

This pull request adds the ability for users to add unlisted professors. This is a fail-safe for when we don't have an up-to-date list of professors for each class. 

This uses [`CreatableSelect` from `react-select`](https://react-select.com/home#creatable) to allow user input in the select dropdown for professors.

### Test Plan 

Tested adding custom professors and verifying that these reviews show up in the `/admin` panel.

| review | admin |
| --- | --- |
| image | image |

### Linter Warnings 

none

### Notes 

For the future, we should just make sure we update the professor lists. Otherwise, we might have some inconsistencies with professor names (e.g., someone might type "Gries" or "Prof Gries" or "David Gries").

### DIFF
--- a/client/src/ui/ReviewForm.tsx
+++ b/client/src/ui/ReviewForm.tsx
@@ -1,5 +1,5 @@
 import React, { useState } from "react";
-import Select from "react-select";
+import CreatableSelect from "react-select/creatable";
 import RatingInput from "./RatingInput";
 import styles from "./css/ReviewForm.module.css";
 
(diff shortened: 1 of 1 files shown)


kungpaogao:
There was some discussion about whether we should just show a "Not listed" option and have them specify in the comment section to avoid the problem listed in the notes.

jack-farley (approved):
Looks good Andrew! Definitely a good feature to have for when professors are not up to date.

juliejeong (commented):
This PR handles the current issue with unlisted professors. It seems good with no issues with functionality. User experience wise, I think there could be some changes that we can make, but considering this is a temporary fix, this seems like a working solution!

michelleli01 (commented):
Looks great Andrew! This new feature will definitely be useful for solving our problem of people selecting random professors when their current professor isn't listed.

kungpaogao:
closing because of #356

PR #351, opened 2022-05-02
Implementing private route

michelleli01:
### Summary 

This pull request is meant to create a protected route that only allows users to access if they are authenticated. For example, users can now only access the `/profile` route if they are already authenticated. Otherwise, if the user tries to access the `/profie` route, they are redirected to the home page.

### Test Plan 
I tested the public and private routes locally.
https://user-images.githubusercontent.com/11303631/166189272-63c89739-1057-4956-a2bc-9bc59256fdfd.mp4


client/src/session-store.ts
```diff
@@ -4,10 +4,27 @@ export class SessionStore {
...
+  }
 
   setPersistent(items: { [key: string]: any }) {
-    Object.entries(items).forEach(([key, value]) => localStorage.setItem(key, JSON.stringify(value)));
+    Object.entries(items).forEach(([key, value]) =>
+      localStorage.setItem(key, JSON.stringify(value)),
+    );
+  }
+
+  isAuthenticated() {
```
AkashAryal, line 17: dupe code

AkashAryal (changes requested):

AkashAryal (approved):

PR #356, opened 2022-05-08
Add "Not listed" option to professors

kungpaogao:
### Summary 

After discussion (see #341), we decided it's better to just have a "Not listed" option for professors. This will make it easier to find/replace old database entries, while still preventing users from selecting incorrect professors. 

Users will also be prompted to include their professor in their review if they select the "Not listed" option.

### Test Plan 

| selected "Not listed" option | normal professor selection | review in admin page |
| --- | --- | --- |
| image | image | image |

### Linter Warnings 

n/a

### DIFF
--- a/client/src/ui/ReviewForm.tsx
+++ b/client/src/ui/ReviewForm.tsx
@@ -88,7 +88,7 @@ export default function ReviewForm({
             );
           }}
           isMulti
-          options={toSelectOptions(professors)}
+          options={toSelectOptions([...(professors || []), "Not listed"])}
           placeholder="Select professors"
         />
         {isSelectedProfessorsInvalid && (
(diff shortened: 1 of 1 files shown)


AkashAryal (approved):
Looks good. I like the not listed option - should make it easier for users to know what to do. I think there is a linter error if you could look at that.

kungpaogao:
> Looks good. I like the not listed option - should make it easier for users to know what to do. I think there is a linter error if you could look at that.

Looks like it's from a different PR - it's the `null` check for `findReviews`:
```
export const findReviews = async (reviewId: string) => {
  try {
    const review = await Reviews.findOne({ _id: reviewId });

    if (review == null) {
    }
  } catch (error) {
    console.log("Error");
  }
};
```

AkashAryal:
Ohh hmm. Could you do a quick fix it in this PR?

PR #362, opened 2022-10-06
Reverting Reviews to Sticky

AnkitLakkapragada:
### Summary 

Work in Progress. 

This pull request is to revert the review form to not be sticky as currently on my_reviews_page, you need to scroll all the way down to the end of the reviews to submit a review of your own. 

### Test Plan 

Testing will compose of using different screen sizes and mobile, (etc). 

image

### Linter Warnings 

No added linter warnings.

### DIFF
--- a/client/src/ui/css/ClassView.css
+++ b/client/src/ui/css/ClassView.css
@@ -10,7 +10,7 @@
 }
 
 .navbar-margin {
-  margin-top: 45px;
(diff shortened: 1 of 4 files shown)


michelleli01 (approved):
Looks good Ankit! This is fix is great for ensuring a good user experience when writing a review :)

PR #363, opened 2022-10-10
Randomized profile bear

michelleli01:
### Summary 

This pull request randomizes the bear profile picture that gets displayed after user logs in.

### Test Plan 
Tested locally.


AnkitLakkapragada (approved):
I like the new bear colors! This change is fun and improves user experience. Good work

PR #365, opened 2022-10-19
Merge Profile Feature :)

michelleli01:
### Summary 
 This PR aims to merge the profile feature that the team has meticulously developed over the last couple semesters. 

### Test Plan 

We have tested each PR into my_reviews_page through our unit tests and manually testing.

### Linter Warnings 


AkashAryal (changes requested):
A few things i found:
* On "http://localhost:3000/profile" the sorting does not seem to work
* Liking is also a bit messed up. The first time you like a post, the number does not change

jack-farley:
> A few things i found:
> 
> * On "http://localhost:3000/profile" the sorting does not seem to work
> * Liking is also a bit messed up. The first time you like a post, the number does not change

Yeah, liking still needs to be addressed. I think its 90% of the way there with the new auth. Given that it currently doesn't work though, I was planning to fix that in a smaller follow up PR since this one is already so big.

AkashAryal:
> > A few things i found:
> > 
> > * On "http://localhost:3000/profile" the sorting does not seem to work
> > * Liking is also a bit messed up. The first time you like a post, the number does not change
> 
> Yeah, liking still needs to be addressed. I think its 90% of the way there with the new auth. Given that it currently doesn't work though, I was planning to fix that in a smaller follow up PR since this one is already so big.

Gotchu. The issue is that if we merge this into master then it will go live into production and we probably shouldn't do that since its not 100% correct. I think you could create a new branch, X, off of my_reviews_page and then do a PR into my_reviews_page and then this PR should include those changes.

jack-farley:
Profile page also gets a little messed up on mobile. Might be worth looking at before we merge.

jack-farley:
> Profile page also gets a little messed up on mobile. Might be worth looking at before we merge.

And the profile bear goes behind the search bar on mobile, I'll take a look at these this weekend.

AkashAryal (changes requested):
Some bugs I found

1. When you write a review and select a professor. Then click on the x (on the professor tag - gray box ) to remove it you get an error. 
EX:
image


The error:
image

2. Sorting by most recent results in a crash in profile page.
image

AkashAryal (approved):
Good job team. Looks good!

PR #369, opened 2022-10-31
Fix profile sorting and hide pending reviews button

michelleli01:
### Summary 

This PR fixes the sorting of reviews on the profile page and also fixes the hide pending reviews button.

### Test Plan 
Tested locally on laptop. Since client keeps making bad requests to server, I just hard coded `http://localhost:8080` to run the code properly. 

### Linter Warnings 

### DIFF
--- a/client/src/ui/Profile.tsx
+++ b/client/src/ui/Profile.tsx
@@ -42,8 +42,7 @@ export default function Profile(imgSrc: any) {
           setVerifiedEmail(res.message);
         }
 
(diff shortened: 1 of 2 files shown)


AnkitLakkapragada (commented):
Looks good! I like how we can hide the pending reviews and it's definitely a very aesthetically pleasing change. Something I was thinking about is if it is better to sort by the latest date first.

jack-farley (approved):
Looks good!

AnkitLakkapragada (approved):
Good work! It's important that it's automatically sorted by most helpful when the page loads so users aren't confused.

PR #370, opened 2022-11-02
fixed bugs from profile page

michelleli01:
### Summary 

There were some bugs on the profile page like sorting by recent reviews and submitting a new review used to both throw errors, but should be fixed now.

### Linter Warnings 


jack-farley (approved):
Looks good!

PR #374, opened 2022-11-09
New review page

michelleli01:
### Summary 

This pull creates new review page including a button instead of the form to save space and lessen users visual strain.

### Test Plan 

new review page
clicking review form


jack-farley (changes requested):
Overall, looks good. If you shrink the window, there's a point where the popup is too big for the window. Maybe that should be fixed in the mobile fix though.

AnkitLakkapragada:
Really useful change that will help the users not have to scroll and create less clutter. I couldn't find any other bugs other than what Jack already commented. Good work!

PR #375, opened 2022-11-09
Fix changing bears bug

michelleli01:
### Summary 

This PR fixes the generated profile bear changing throughout the site.

### Test Plan 

maintain bear image
maintain bear image 2


jack-farley (changes requested):
Somehow I get to a point where the profile bears don't load. The server shows this error:

`Error: ENOENT: no such file or directory, stat '/Users/jackfarley/Dev/school/dti/course-reviews-react-2.0/client/build/index.html'`

Not sure exactly why its happening, could be local on my end.

PR #377, opened 2022-11-10
Raffle

jack-farley:
This PR adds functionality on the admin side to choose a raffle winner based on a given start date.


michelleli01 (approved):
Looks good! This will make picking raffle winners more convenient.

PR #396, opened 2023-09-11
Transition to new DTI logo & Begin Frontend refactoring 

wizhaaa:
# Summary 

This PR does the following: 

- updates logo to new name 
- begins frontend refactoring
- `App.tsx` migrated to `Home.tsx` (in modules folder) 
- Background image logic refactored 
- Using functional arrow based components 
- es5 -> es6 

## Test Plan 

- [ ] Checked page with different seasons & times. 
- [ ] Logged in, logged out.
- [ ] Logo changes to white text at night? (manual and actual night-time IRL) 

### Images
Before
image
After
image

# Potential Future To-Dos 

- [ ] Migrate to `react-router-v6` 
- [ ] Migrate `CRA` to `vite` 
- [ ] `tsconfig.paths` : for path aliasing (no more `../../../..`). Attempted but no luck so far - kept breaking, could not resolve. 
- [ ] Refactor front-end to following schema: 
- [ ] Delete unused files, variables, imports (including the `App.tsx` file => will work on in next PR) 
- [ ] change `master` to `main` branch 
- [ ] update read me 

client
├── src
│   ├── assets 
        └── imgs
│   └── log.txt
└── modules
    └── PageName
        └── Components
        └── Styles
                    index.tsx
├── App.tsx (routers & global
├── index.tsx (react router dom)


client/src/modules/Home/Components/Home.tsx
```diff
@@ -0,0 +1,119 @@
...
+  useEffect(() => {
+    setBackground();
+    if (time === "night") {
+      setDTILogo(DTIWhiteLogo);
+    }
+  }, [time]);
+
+  /** Displays "sign in" or profile bear picture */
+  const NavButton = () => {
+    if (netId) {
```
kembayeb, line 66: assuming that `netId` populates if user is signed in? Why not use `isLoggedIn` bool here for the if statement?

client/src/index.tsx
```diff
@@ -2,17 +2,18 @@ import "bootstrap/dist/css/bootstrap.min.css";
 import "./index.css";
 
 import React from "react";
-import { render } from "react-dom";
+import {render} from "react-dom";
```
michelleli01, line 5: Why remove this white space? Is this after linting with the prettier extension?

client/src/modules/Home/Components/Home.tsx
michelleli01: Agreed, this was code that was developed after the isLoggedIn bool was created.

client/src/ui/css/App.css
```diff
@@ -170,15 +164,10 @@
...
-/* I believe this is only used in recentreview.jsx so will remove soon TODO */
-.container {
-  background-color: rgba(255, 255, 255, 0);
-}
-
 .sign-in-button {
   position: absolute;
-  right: 80px;
-  top: 50px;
+  right: 30px;
```
michelleli01, line 169: Is there a reason for this sign in button change?

michelleli01 (commented):
Look great! Thanks for all the great changes Will! The refactoring looks great so far.

client/src/modules/Home/Components/Home.tsx
wizhaaa: Thanks for the catch. To be honest, I just copied and pasted the code from before but just refactored the logic behind it - did not really change this NavButton functionality-wise. I just updated it to use `isLoggedIn`. To be honest, I want to restructure the entire `useAuthOptionalLogin()` to return an object or a module with exports, not an array - which would also solve the unused variables warning. 

Will most likely do this in a next/future PR as long as it isn't too much.

client/src/index.tsx
wizhaaa: Yup, I think prettier was broken or something for me. I added a prettier config file for the front end. 

=> (pressed save and now it also just fixed formatting for other files too)

client/src/ui/css/App.css
wizhaaa: On mobile, the sign-in button looked jank (like near the middle of the screen) rather than flushed to the side. 

I did this as a sort of temporary fix as I think it would look better if the sign in buttons were consistently flushed to the side.

I had the right for 30px as the margin-right of the search bar was also 30px (but doesn't exactly line up using 30px right). 

Changed it to like ~15px right. 

**Before:**
image

**After:**
image

client/src/modules/Home/Components/Home.tsx
michelleli01: I agree, the `useAuthOptionalLogin()` is not implemented too well. This is a fix we can definitely make an add onto the list of tasks. I want to move away from Google login as it seems to be not as well maintained since Google seems to prefer just the use of Firebase authentication. We can take a look at other dependencies such as `Passport.js`. I'll do some research.

client/src/index.tsx
michelleli01: Thanks!

client/src/ui/css/App.css
michelleli01: Oh yes perfect! Thank you so much! This was one of the bugs I wanted to fix 👍💯

michelleli01 (approved):
Looks great Will thank you so much!

PR #398, opened 2023-09-13
Client Side Restructuring

wizhaaa:
# Summary 

This PR mainly restructures the frontend code. 

Just moved the files into more neat and tidy modules rather than one giant UI folder and one giant CSS folder for all components. 

Updated the file import paths, CSS, etc. 

Main structure: 
Each "page" (that has a route) will have its own folder under `modules`. 

- index page to export to the router. 
- Components folder to hold all relative components to the page. 
- Styles folder to hold all relative `css` files (or if just one `css` file - not inside a Styles folder) 

# Testing 

Tested with most features - home page, search bar, making a review, profile page, etc... 

### Main Structure: 
image
image
image

# Concerns & Future topics to research

### `Accordian.tsx` under UI. 
@michelleli01 - Will we need to use this? Or can we delete it and build a new accordion if needed? It is imported into the `Stats/Admin` module, but is commented out and unused. Thinking of just deleting and thus no longer having a UI folder. 

### Js & JSX files, Class components
Some files are JSX instead of TSX. Some use class-based components instead of functional components. 
Want to update this in the future ~ more efficient if someone will be working on the feature/files to update it along with implementing their feature. 

### Using `.css` or `.module.css` 
Do we want to use one or the other? Currently use both. I do feel `.css` is more simple. But, I am not fully aware of the tradeoffs between them. 

### Want to refactor `useAuthOptionalLogin()` 
Lots of linting problems such as unused components because we return an array of values - which we do not always use. Ideally, I want to make it a function that returns an object or a module with exportable values rather than an array.


kembayeb (commented):
@wizhaaa i like this refactor, nice and organized! I think you might've accidentally deleted the homepage images tho, this is what i see locally
Screen Shot 2023-09-21 at 5 37 14 PM

wizhaaa:
> @wizhaaa i like this refactor, nice and organized! I think you might've accidentally deleted the homepage images tho, this is what i see locally Screen Shot 2023-09-21 at 5 37 14 PM


Quite odd - works on my end T.T.

Will just change it back to be safe.

wizhaaa:
Which I realize now is because it fetches the old background from the cache first actually

michelleli01 (approved):
This looks great Will! The restructuring is much more intuitive in terms of frontend files. I would say for the `Accordian.tsx` file, that can be deleted. It's not being used and is deprecated. We could build a new accordian since that's being used for statistics displayed on the Admin page I believe. But, honestly I'm think we should just tap into Google Analytics more instead having manually displaying stats on the Admin page. There might be more deprecated files in terms of old `.jsx` vs `.tsx` files I'll have to look deeper into it.
In terms of using `.css` vs `.module.css`, if you read the Developer Documentation Dev Workflow, I explain why we should use `.module.css` over `.css`. But the TLDR is that `.module.css` helps localize classes instead of having them global. This allows for portable and reusable CSS files. They avoid unintended cascading. 
Refactoring `useAuthOptionalLogin()` may be beneficial for the future for sure. I would say we're currently using Google auth, which has some deprecated functions. This may need more thinking behind it in terms of whether we stick to Google auth's module or migrate somewhere else. The Google auth library has the problem of inserting `null` values to new students in terms of First and Last names. I think this is because of some changes made to the library, but I haven't done extensive research.

wizhaaa:
Thanks, Michelle - I removed the accordion. The admin page front-end might just be better if done from scratch design-wise yep. 

I will look into it / or most likely assign the auth situation to a developer (& help them through it) 

Got it, I'll also look into the CSS modules documentation you wrote + learn more about it too.

PR #403, opened 2023-09-26
Typography Change & Tiny CSS Reset

wizhaaa:
# Summary 

According to Maywa's designs - changing the font to SF Pro (apple-system font which falls back to sans-serif). May actually need to install it as a web font if it does not display the same on a Windows machine. 

 - CSS Reset (some global CSS changes to make future styling easier 
 - Remove the Mobile Metrics view (buggy right now) - also the course reviews is not sticky rn 
 
# Comments for the future 
- CSS files are sort of messy and have a lot of unnecessary lines. ( Looks like copied from Figma ) ie. font is constantly re-defined when not needed. 
- I would love to redo the CSS Styling in the future ( new dev task o.o ) - get rid of any bootstrap classes so we don't have to rely on it anymore (has a lot of built-in margins/paddings/spacing makes it hard to style properly.)
- The sign-in button is blocked by the search bar on mobile (new dev task) yay!
- Review page modal full-screen (new dev task yay)

# Testing 
### It looks solid to me so far
image

### DIFF
--- a/client/src/index.css
+++ b/client/src/index.css
@@ -29,7 +29,7 @@ body {
   padding: 0;
   margin: 0;
   font-size: 14px;
-  line-height: 1;
+  /* line-height: 1.2; */
 }
 
 /* HTML5 display-role reset for older browsers */
--- a/client/src/index.css
+++ b/client/src/index.css
@@ -3,6 +3,22 @@
   --semi-bold-weight: 600;
 }
 
+* {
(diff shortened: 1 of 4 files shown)


michelleli01 (approved):
Changes look great Will! Thanks for implementing Maywa’s font change. It definitely looks a lot better now!

PR #405, opened 2023-10-01
Mobile profile page

harrisonchin2523:
### Summary 

These changes are following the Figma design (https://www.figma.com/file/P6CKTAnkypxYP0PHBKhDKd/FA22?type=design&node-id=203-684&mode=design&t=BWzq5h1DHEHGhUQT-0). Let me know if you see anything wrong or see any changes I should make as I familiarize myself a bit more with the codebase. 

Screenshot 2023-10-01 at 7 29 49 PM


wizhaaa (commented):
Some initial comments about the CSS & Styling so far

leihelen (commented):
I think everything on the no reviews mobile page so far looks really good and matches the figma so just keep up what you're doing! I think completing this PR will be really beneficial since it makes the mobile version of the website easier to use.

wizhaaa (commented):
The mobile view for users with no reviews definitely looks really good! Just have a few concerns about the overall responsiveness of the rest of the page & when a student does have past reviews.

image

~ This is around the screen width of 912px 
I know this is a bit out of the scope of the "mobile" profile page - but ideally, we want the site to be as responsive as possible. 

Some possible solutions for this are to either make the "My Dashboard" white section full width as well and decrease the spacing between that and the "My Reviews" Section. 


Same around ~1000px wide: 
image
Text clipping here as well

wizhaaa:
Also - I know a lot of this is going beyond a simple mobile view of the profile and more just kind of completely re-doing it instead. I think at this point, the mobile page is "servicable" - if you rather work on it on another PR or work on it after the tags feature. You can also do that. Although, I think it would be best to keep it all in one PR when you already changing the associated code. 

Other comments (Sorry - I forgot to mention and I closed my tab with my draft comment): 

### Editing
- There's still "flag icon" on the reviews on our profile page. Designs show a pencil icon to edit 
- Edit functionality is optional -> Could either have a placeholder icon button and say "editing reviews is not currently supported" pop-up. 
- Editing functionality will be another PR for sure 

### Review Cards
The styling of each card is not entirely faithful to the designs either: 
image
image

### Minor CSS bugs
There seems to be a margin or something here (right-margin) that causes it to not full align and not fully take up 100% width of its container. 

AND the date is taking up too much width which is squeezing the helpful container and causing it to shrink and be cut off. 

image

wizhaaa:
# Code Reccs: 
- Make a new component (under Profile folder) called `ReviewCard` instead of using CourseReview component (should probably also be renamed in future) -> pass props in; be able to style different if pending + approved 
- Try to delete all bootstrap classes (col, row, etc). It causes a lot of bugs in CSS and makes it behave in ways you wouldn't expect. 
- Perhaps modulate each section of the profile page. Right now, it does look like one big essay of code. Like: 
```
return  
<navbar> 
<userinfo> 
<pendingreviews> 
<approved reviews> 
<etc.> 
```

where these other components are in another file under Profile

client/src/modules/Profile/Component/PendingReviews.tsx
```diff
@@ -0,0 +1,47 @@
...
+      </div>
+      <div className={hide === false ? styles.reviewCard : ''}>
+        {hide === false ? (
+          <CourseReviews
+            reviews={pendingReviews}
+            isPreview={false}
+            isProfile={true}
+          />
+        ) : (
+          <p></p>
```
kembayeb, line 40: Looks good Harrison! It seems like this p tag isnt being used

michelleli01: This is fine since it just signifies to display nothing is `hide === true`

michelleli01 (approved):
Looks great Harrison! Thanks for all the hard work. Changes seem to be responsive and I like the use of CSS modules 👍

PR #406, opened 2023-10-02
Add documentation to Profile.ts

leihelen:
I added comments to some of the functions in Profile.ts to help with documentation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -116,6 +116,8 @@ You will probably need to run this for the `classes`, `subjects` and `reviews` c
 - **Michelle Li** - Developer Advisor
 - **Harrison Chin** - Developer
 - **Samuel Yeboah** - Developer
+- **Helen Lei** - Developer
+- **Andrea Thia** - Designer
 
 ### 2022 - 2023
 
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -50,7 +50,7 @@
 </head>
 
 <body>
-  <div id="render-target" class="full-height">Please enable javascript to view this page.</div>
+  <div id="render-target" class="full-height"></div>
 </body>
 
 </html>
--- a/client/public/index.html
+++ b/client/public/index.html
@@ -50,7 +50,7 @@
 </head>
 
 <body>
-  <div id="render-target" class="full-height"></div>
+  <div id="render-target" class="full-height">Please enable javascript to view this page.</div>
 </body>
 
 </html>
--- a/server/endpoints/Profile.ts
+++ b/server/endpoints/Profile.ts
@@ -9,10 +9,14 @@ export interface NetIdQuery {
   netId: string;
 }
 
+// The type of a query with a token
 export interface ProfileRequest {
   token: string;
 }
 
+/**
+ * [getStudentEmailByToken] returns a student email if the given token is valid
+ */
 export const getStudentEmailByToken: Endpoint<ProfileRequest> = {
   guard: [body("token").notEmpty().isAscii()],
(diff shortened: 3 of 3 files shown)


client/public/index.html
```diff
@@ -50,7 +50,7 @@
 </head>
 
 <body>
-  <div id="render-target" class="full-height"></div>
+  <div id="render-target" class="full-height">Please enable javascript to view this page.</div>
```
michelleli01, line 53: Did you mean to make this change?

README.md
```diff
@@ -116,6 +116,8 @@ You will probably need to run this for the `classes`, `subjects` and `reviews` c
 - **Michelle Li** - Developer Advisor
 - **Harrison Chin** - Developer
 - **Samuel Yeboah** - Developer
+- **Helen Lei** - Developer
```
michelleli01, line 119: A minor thing: maybe specify that this PR is also adding people to the README or make it a separate PR

server/endpoints/Profile.ts
```diff
@@ -9,10 +9,14 @@ export interface NetIdQuery {
   netId: string;
 }
 
+// The type of a query with a token
 export interface ProfileRequest {
   token: string;
 }
 
+/**
+ * [getStudentEmailByToken] returns a student email if the given token is valid
```
michelleli01, line 18: This is my bad for not specifying the requirements of a doc string. But, ideally we would have something like this:
```
/**
   * Returns true if [insert something here] matching the email if the given token is valid and false otherwise.
   * This method authenticates the user token through the Google API.
   * @param token: google auth token
   * @requires that you have a handleVerifyError, like as follows:
   * verify(token, function(){//do whatever}).catch(function(error){
   */
```

Something similar to the code I wrote above to specify parameters, return types, etc. Granted this file doesn't do that well or consistently hence the new documentation requirement

server/endpoints/Profile.ts
```diff
@@ -36,7 +40,7 @@ export const getStudentEmailByToken: Endpoint<ProfileRequest> = {
 };
 
 /**
- * Counts the number of reviews made by a given student id.
+ * [countReviewsByStudentId] returns the number of reviews made by a given student id.
```
michelleli01, line 43: See above.

michelleli01 (changes requested):
Great work! Sorry about the unclear instructions for the dev portfolio assignment. I left some comments :)

server/endpoints/Profile.ts
```diff
@@ -9,10 +9,14 @@ export interface NetIdQuery {
...
+// The type of a query with a token
 export interface ProfileRequest {
   token: string;
 }
 
+/**
+ * [getStudentEmailByToken] returns a student email if the given token is valid
+ */
 export const getStudentEmailByToken: Endpoint<ProfileRequest> = {
   guard: [body("token").notEmpty().isAscii()],
```
michelleli01, line 21: I hint I wanted to add especially since the method CU Reviews uses to create APIs is unique and kinda weird, if you want to check the parameters taken in, check the `guard` variable and take a look at want is required. So for example here in this `getStudentEmailByToken` function, a `token` variable is required and is being checked to see if it's empty and an ASCII value. This is ensure types are rigid and there isn't much room for error.

wizhaaa:
I think you may have accidentally created this branch off of your other branch in your other PR (adding to README branch)

Next time, make sure to switch back to the master branch, run git pull to get any updates off of the remote repository, and then create a new branch from the master branch.

PR #411, opened 2023-10-22
Shrink search bar and sign-in button

leihelen:
I made the width of the search bar and sign-in button smaller when the page is narrow so that the sign in button isn't blocked on mobile.


client/src/modules/Globals/Styles/NavBar.css
```diff
@@ -10,7 +10,7 @@
 
   .navbar-searchbar-container {
     margin-left: 21px;
-    margin-right: 70px;
+    margin-right: 150px;
```
wizhaaa, line 13: In the future, we should try to stay away from hard-coding margins - not very responsive compared to relative widths.

wizhaaa (changes requested):
Hi Helen! ~ Thanks for starting this feature! 

It definitely looks great when there is no User signed in. 

However when there is a user signed in, there is too much spacing (margin that was hard coded in). 

See the pictures below: 

### User Signed In - too much spacing 
image

### User Logged out - looks great. 
image

## Solution? Ranking from least to most work:
1. Replace the Sign-In Button with a grey-scale profile picture. Change the `margin` back to a lower value. 
2. OR If you can somehow conditionally apply a different CSS class (to change a different margin value) conditionally on if a user is logged in (lower margin) or not (keep current margin value) 
3. OR [hardest _but_ the most rewarding _and_ improves future code quality] Remove the hard-coded margin, remove the `col` bootstrap classes on the SearchBar component + TBH just restart the entire CSS for the SearchBar Component & Search Bar/NavBar Container from scratch and build it up from the ground up. 
- #3 is tricky because TBH the SearchBar is such wonky code (still uses legacy class components instead of functional) - which if you are able to decipher would be a nice thing to tackle/learn if you wanted to try it 

I can help you during TeamTime/WorkSess if you need any help or guidance as well!

client/src/modules/Globals/Styles/NavBar.css
wizhaaa: BUTTTT like probably easiest approach & solution for this case

PR #412, opened 2023-10-22
Add profile documentation

leihelen:
changed the documentation in Profile.ts

### DIFF
--- a/server/endpoints/Profile.ts
+++ b/server/endpoints/Profile.ts
@@ -15,7 +15,7 @@ export interface ProfileRequest {
 
 /**
  * This method authenticates the user token through the Google API, returns true
- * if [token] matches the email if the given token is valid and false otherwise
+ * if [token] matches the email if the given token is valid and false otherwise.
  * @param token: google authentication token that is checked to see if it is not
  * empty and an ASCII value
  * @requires that you have a handleVerifyError, like as follows:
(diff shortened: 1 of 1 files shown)


harrisonchin2523 (commented):
Hey Helen, nice job addressing the changes that Michelle requested. I noticed that you just have some things in the docstrings that might not be totally correct. For some of the docstrings, you state that the function will throw an error "if there are none", but this isn't necessarily accurate. For example. getTotalLikesByStudentId will just return a message with totalLikes = 0 and not throw an error. Did you mean if the findOne methods didn't find a document, then they would return null, or it would error if let's say the netId did not exist?

server/endpoints/Profile.ts
```diff
@@ -60,7 +72,11 @@ export const countReviewsByStudentId: Endpoint<NetIdQuery> = {
 };
 
 /**
- * [getTotalLikesByStudentId] returns the total number of likes a student has gotten on their reviews
+ * Returns the number of likes that match the reviewIds associated with a given
+ * netId and error if there are none
+ * This method counts the total number of likes received by a student.
```
kembayeb, line 77: hey Helen, nice work! I think here we should change `"This method counts the total number of likes received by a student."` to  `"This method calculates the total number of likes a student has received on all their reviews."` to be a little more clear

harrisonchin2523 (changes requested):
Just being a little bit nitpicky here, but it seems like some of the docstrings have redundant information. For example, countReviewsByStudentId states what the function does twice in the first three lines.

PR #415, opened 2023-11-01
create clusters based on most similar course descriptions

kembayeb:
### Summary 

Created clusters based on the similarity of course descriptions, using a library called genism to do those calculations for us. 

Clusters are based on all courses offered in FA23.

Each cluster has 50 coursed inside, or lightly more if an outlier was merged inside. Also currently have 87 clusters.

NOTES
currently courses that don't have a course description aren't accounted for which is  97 courses

### DIFF
--- a/course-clusters/clusters.py
+++ b/course-clusters/clusters.py
@@ -1,54 +1,12 @@
-import string
 from gensim.models.doc2vec import Doc2Vec, TaggedDocument
-from load_course_info import loadCourseDesc, getSubjects
-
-subjects_url = "https://classes.cornell.edu/api/2.0/config/subjects.json?roster=FA23"
-subjects = getSubjects(subjects_url)
-
-urls = [f"https://classes.cornell.edu/api/2.0/search/classes.json?roster=FA23&subject={sub}" for sub in subjects]
-
-def fetchDescriptions(urls):
-    descriptions = []
-    for x in urls:
-        descriptions += loadCourseDesc(x)
-    print(len(descriptions))
-    return descriptions
-
-
-course_descriptions = fetchDescriptions(urls)
-def preprocess_text(text):
-    """ Tokenization and preprocessing with custom stopwords"""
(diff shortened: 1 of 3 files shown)


course-clusters/clusters.py
```diff
@@ -0,0 +1,116 @@
+import string
+from gensim.models.doc2vec import Doc2Vec, TaggedDocument
+from load_course_info import loadCourseDesc, getSubjects
+
+subjects_url = "https://classes.cornell.edu/api/2.0/config/subjects.json?roster=FA23"
+subjects = getSubjects(subjects_url)
```
michelleli01, line 6: Can we move this section into a new file and import the subjects variable.

course-clusters/clusters.py
```diff
@@ -0,0 +1,116 @@
...
+
+def fetchDescriptions(urls):
+    descriptions = []
+    for x in urls:
+        descriptions += loadCourseDesc(x)
+    print(len(descriptions))
+    return descriptions
+
+
+course_descriptions = fetchDescriptions(urls)
```
michelleli01, line 18: Can we also move this to a separate file? We can add the `getSubjects` and `fetchDescriptions` functions into a separate `scrapper.py` file.

course-clusters/clusters.py
```diff
@@ -0,0 +1,116 @@
...
+
+def removeStopwords():
+    "Removes stopwords for all the descriptions "
+    preprocessed = []
+    for desc in course_descriptions:
+        if desc:
+            preprocessed.append(preprocess_text(desc))
+    return preprocessed
+
+preprocessed_descriptions = removeStopwords()
```
michelleli01, line 44: Let's also move this preprocessing out. Separate the preprocessing/web scrapping with the actual model itself. Can we also create some persistent objects to store the web scrapping data and the preprocessing so we don't always have to do that?

michelleli01 (changes requested):
It looks great Kemba! I'm so excited to see progress with the recommendations feature. I left some minor comments about separating certain sections within your `clusters.py file

PR #416, opened 2023-11-05
Adding review tags

harrisonchin2523:
### Summary 

Currently, working on adding tags to the reviews. The current implementation is only visual and the buttons on the review form do not have any functionality. Will be working on adding functionality to the buttons so that the reviews do not just display all the possible tags like they do right now.

### Test Plan 

Screenshot 2023-11-05 at 3 47 20 PM
Screenshot 2023-11-05 at 3 46 12 PM


client/src/modules/Globals/tag_names.ts
leihelen, line 1: I think the tags look really good so far and this is a super helpful feature since it can help people get important information for each class easily! Maybe you could add some more tag options as well, such as a tag for whether it is attendance mandatory or if it's discussion or lecture based.

PR #420, opened 2023-11-15
fix sign-in button blocked on mobile

leihelen:
Shrank sign-in button and added conditional css class to make searchbar wider if user is logged in. I made a new branch/pull request because my cla got messed up on the last one. 

Screenshot 2023-11-15 at 5 39 17 PM
Screenshot 2023-11-15 at 5 39 04 PM


client/src/modules/Globals/Navbar.tsx
```diff
@@ -68,12 +68,12 @@ export default function Navbar({ userInput }: NavbarProps) {
...
           <img
             src="/logo.svg"
-            className="img-fluid scale-logo-navbar"
+            className="scale-logo-navbar"
             alt="CU Reviews Logo"
           />
         </a>
       </div>
-      <div className="col navbar-searchbar-container">
+      <div className={`col navbar-searchbar-container${isLoggedIn ? "-wider" : "-not"}`}>
```
michelleli01, line 76: More so for convention, instead of `-not` , you can just leave the `className` as an empty string if the user is not logged in. So, if `isLoggedIn` is `true`, then the `className` will be `navbar-searchbar-container`, but is it's `false`, then the `className` will be `navbar-searchbar-container`. Also,

nit: can we change the `className` from `navbar-searchbar-container-wider` to `navbar-searchbar-container-profile`, or something more related to why this `className` is being used?

client/src/modules/Globals/Styles/NavBar.css
```diff
@@ -10,7 +10,17 @@
 
   .navbar-searchbar-container {
     margin-left: 21px;
-    margin-right: 70px;
+    margin-right: 150px;
+  }
```
michelleli01, line 14: Since this class isn't being used with your current implementation, you can delete this. Or, if you decide to implement the changes I suggested above (completely optional) then you can delete `navbar-searchbar-container-not`.

client/src/modules/Profile/Styles/App.css
```diff
@@ -29,6 +51,28 @@
...
+    font-family: Source Sans Pro;
+    font-style: normal;
+    font-weight: 600;
+    font-size: 18px;
+    line-height: 23px;
+    text-align: center;
+    letter-spacing: -0.22px;
+
+    color: #000000;
+  }
```
michelleli01, line 75: For changing the CSS with media tags, you only need to add the line that is being changed. In this case, you only need to include `width: 100px`, but it's a bit strange to go from percentages to pixels, so could this be changed to a percentage as well? This makes the file shorter and more consistent especially since `App.css` is quite 😢

michelleli01 (changes requested):
Great work on your first official PR! I just left some minor nit picks.

michelleli01:
> Shrank sign-in button and added conditional css class to make searchbar wider if user is logged in. I made a new branch/pull request because my cla got messed up on the last one.

Oh also, can you include screenshots of what your changes look like now? Like the difference between the searchbar with the sign in button and the profile bear.

michelleli01 (approved):
Looks great Helen! Thanks for making those changes so quickly, it looks good on my end. Great work!!! 😄

PR #433, opened 2024-02-20
React-Scripts Update to 4.0.3

wizhaaa:
# Summary 

After long-awaited pondering and bug-fixing, I have finally successfully (on my local computer at least) migrated react-scripts from 3.4.1 to 4.0.0. 

- also removes old dependencies and deletes them from admin & other various pages 
- reroute home page background image to assets

## Testing

You may have to delete the `node_modules` under the root directory, `client` directory, and `server` directory and run `yarn` (in the root) to install the dependencies again. 

### Thoughts 

I wanted to migrate to the latest (5.0.1) but there seemed to be a lot of breaking changes on the server-side code (especially linting and jest). 

I tried to solve it - but I think we may have to update jest like up like 5 major versions - lots of rewriting tests?

Or nuke and go to vite.


leihelen (commented):
Everything works well on my local machine as well and I like how you preserved the structure of the code! This is really amazing work and migrating to a newer version of react-scripts will greatly improve CU Reviews. Touching on your thoughts about migrating to 5.0.1, I think it might be better to look more into vite rather than continuing with react-scripts. Even though it would definitely be a lot of work, I think vite would have a lot of benefit in the long run and be more worth it for the amount of effort it takes compared to migrating to 5.0.1.

leihelen (approved):
Everything works well on my local machine as well and I like how you preserved the structure of the code! This is really amazing work and migrating to a newer version of react-scripts will greatly improve CU Reviews. Touching on your thoughts about migrating to 5.0.1, I think it might be better to look more into vite rather than continuing with react-scripts. Even though it would definitely be a lot of work, I think vite would have a lot of benefit in the long run and be more worth it for the amount of effort it takes compared to migrating to 5.0.1.

PR #435, opened 2024-03-03
Helenlei/ai gpt costing

leihelen:
Script that can be used to get the numbers needed to calculated the total costs for creating "cornellian says" summaries, depending on which model we use. For example, below are the results for if we wanted to create summaries for all courses that have at least 3 reviews. 
Screenshot 2024-03-02 at 8 36 05 PM


qiandrewj (commented):
This looks great, Helen! Code is streamlined and I like that we have different options to only look at courses with a minimum number of reviews - it'll be a huge help as we look into OpenAI costing. The one thing I can point out is in the '/costing' route it would be helpful to get the minimum from the request body and pass it in to the function. Great work overall :)

jacquelinecai:
Great work Helen! The structure of the code is very easy to follow. I think something that might also help with the OpenAI costing would be to look at the average number of distinct words in each review.

wizhaaa:
This is amazing Helen - great to see you getting the pieces connected together and running the functions on our staging database. 

Here are some more reference points which you can use to provide & calculate more concrete prices from https://openai.com/pricing

(4 characters = 1 token) and GPT-4 and GPT-3.5 turbo have some pretty drastic price points. 

image

Currently discussing production-level database access with Sarah right now - looking like we may be able to get temporary read access first.

PR #439, opened 2024-03-24
Helenlei/costing admin page

leihelen:
Created a dashboard summary of OpenAI API costs that can be reached at '/admin/costing'. Can choose on the dropdown menu a minimum number of reviews to run the summary on and then will generate the costing estimates for that minimum. 
Screenshot 2024-03-23 at 9 33 13 PM

### DIFF
--- a/client/src/App.tsx
+++ b/client/src/App.tsx
@@ -6,6 +6,7 @@ import { Home } from './modules/Home'
 import { Course } from './modules/Course'
 import { Profile } from './modules/Profile'
 import { Admin } from './modules/Admin'
+import { AdminCosting } from './modules/AdminCosting'
 
 /* Importing Helpers */
 import { Login } from './modules/Login'
@@ -26,16 +27,17 @@ import './index.css'
 const App = () => {
   return (
     <Router>
-        <Switch>
-          <Route exact path="/" component={Home} />
-          <Route exact path="/admin" component={Admin} />
-          <Route exact path="/course/:subject/:number" component={Course} />
-          <Route exact path="/auth" component={AuthRedirect} />
-          <Route exact path="/login" component={Login} />
-          <Route exact path="/results/:type/:input" component={Results} />
-          <Route exact path="/profile" component={Profile} />
-          <Route component={NotFound} />
-        </Switch>
+      <Switch>
+        <Route exact path="/" component={Home} />
+        <Route exact path="/admin" component={Admin} />
+        <Route exact path="/admin/costing" component={AdminCosting} />
+        <Route exact path="/course/:subject/:number" component={Course} />
+        <Route exact path="/auth" component={AuthRedirect} />
+        <Route exact path="/login" component={Login} />
+        <Route exact path="/results/:type/:input" component={Results} />
+        <Route exact path="/profile" component={Profile} />
+        <Route component={NotFound} />
+      </Switch>
     </Router>
   )
 }
--- a/client/src/modules/AdminCosting/Components/AdminCosting.tsx
+++ b/client/src/modules/AdminCosting/Components/AdminCosting.tsx
@@ -1,0 +1,231 @@
+import React, { useEffect, useState } from 'react'
+import { Redirect } from 'react-router-dom'
+
+import axios from 'axios'
+
+import { Review } from 'common'
+
+import { useAuthMandatoryLogin } from '../../../auth/auth_utils'
+
+import UpdateReview from '../../Admin/Components/AdminReview'
+import Stats from '../../Admin/Components/Stats'
+import RaffleWinner from '../../Admin/Components/RaffleWinner'
+
+/** Admin Page
+ * Approve new reviews, see stats, and import new semester courses & Profs.
+ */
+export const AdminCosting = () => {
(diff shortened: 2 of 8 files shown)


client/src/modules/AdminCosting/Components/AdminCosting.tsx
```diff
@@ -0,0 +1,138 @@
...
+          <select
+            id="minReviews"
+            value={minimumReviews}
+            onChange={(e) => {
+              const value = Number(e.target.value);
+              setMinimumReviews(value);
+              getCostingData(value, setTotalTokens);
+            }}
+          >
+            <option value="100000">x</option>
```
qiandrewj, line 88: Maybe instead of setting the initial or "empty" state as an arbitrarily high limit for minimum reviews, it could just default to the lowest minimum number of reviews (3).

client/src/App.tsx
```diff
@@ -6,6 +6,7 @@ import { Home } from './modules/Home'
 import { Course } from './modules/Course'
 import { Profile } from './modules/Profile'
 import { Admin } from './modules/Admin'
+import { AdminCosting } from './modules/AdminCosting'
```
qiandrewj, line 9: One thing I'm not quite sure about, but could be a more clean way to separate the admin page and its functionality from the rest of the app - maybe Admin could contain a router of its own so that new pages to Admin are not all added to the overall app.

client/src/modules/AdminCosting/Components/AdminCosting.tsx
```diff
@@ -0,0 +1,138 @@
...
+      confirmAdmin()
+    }
+  }, [isLoggedIn, token, isAuthenticating])
+
+  // Takes in a minimum number of reviews selected from a dropdown menu.  
+  // Gets data from all courses that have at least that number of reviews.
+  function getCostingData(min: number, setTotalTokens: React.Dispatch<React.SetStateAction<number>>) {
+    console.log('loading costing data...');
+
+    axios.post('/ai/costing', { min })
```
qiandrewj, line 54: It may be more clear what this route is related to if it was changed to something like /admin/aicosting

qiandrewj (commented):
Overall, the implementation of the costing page was really well done, and used a clean UI that I thought was perfect for getting all of the behind-the-scenes information from the admin page -- I'll be drawing from that for the future. Only some small changes I left throughout the code to keep the codebase clean, but this will be useful for us moving forward!

client/src/modules/AdminCosting/Components/AdminCosting.tsx
```diff
@@ -0,0 +1,138 @@
...
+    async function confirmAdmin() {
+      const res = await axios.post(`/api/tokenIsAdmin`, {
+        token: token,
+      })
+
+      if (res.data.result) {
+        setIsAdmin(true)
+      } else {
+        setIsAdmin(false)
+      }
```
jacquelinecai, line 40: I think you could condense this into `setIsAdmin(res.data.result)` for easier readability.

client/src/modules/AdminCosting/Components/AdminCosting.tsx
```diff
@@ -0,0 +1,138 @@
...
+          <p>$60/1M tokens output</p>
+        </div>
+        <div className="gpt3">
+          <h1>${gpt3cost}</h1>
+          <h2>GPT-3.5 cost</h2>
+          <p>$0.50/1M tokens input</p>
+          <p>$1.50/1M tokens output</p>
+        </div>
+
+        <div className="other-numbers">
```
jacquelinecai, line 114: Instead of "other-numbers", this class name could be "review-stats" or "review-numbers" to better represent the information rendered here.

client/src/modules/AdminCosting/Components/AdminCosting.tsx
```diff
@@ -0,0 +1,138 @@
...
+          <p>Number of average characters per review: {avgChars}</p>
+          <p>Number of average tokens per review: {avgTokens}</p>
+        </div>
+      </div>
+    )
+  }
+
+  function adminLogin() {
+    if (loading) {
+      return <div>Loading...</div>
```
jacquelinecai, line 129: Once you pull Andrew's PR for the loading screen, replace "Loading..." with the loading spinner component using `return <Loading />`

server/src/ai/routes.ts
```diff
@@ -0,0 +1,31 @@
...
+      });
+    }
+
+    return res.status(200).json({
+      message: 'Retrieved all pending reviews',
+      result: costing,
+    });
+  } catch (err) {
+    return res.status(500).json({ error: `Internal Server Error: ${err}` });
+  }
```
jacquelinecai, line 27: I would move around some of these status responses to keep it consistent with the rest of the codebase
```suggestion
  try {
    const costing = await minReviewsCosting(min);
    if (!costing) {
     return res.status(500).json({ error: `Internal Server Error` });
    }

    return res.status(200).json({
      message: `Retrieved all pending reviews for a minimum value of ${min}`,
      result: costing,
    });
  } catch (err) {
    return res
        .status(400)
        .json({ error: `No reviews found` });
  }
```

PR #441, opened 2024-04-14
"Cornelians Say"

leihelen:
This PR works on the "Cornellians say" feature. I connected to OpenAI and added endpoints to get all courses with at least a minimum number of reviews, get text from all reviews in a course, and then generate a summary using this text. Right now the code doesn't work when running the server because I think OpenAI might require at least node 18 and/or typescript 4.5, however I think if you comment out all the OpenAI code, run the backend, and then comment it back in it should work. I also created a new node project that uses node 20 and typescript 5.4.5 and it worked there. 
Here are 2 example summaries I generated from a course with at least 5 reviews and a course with at least 10 reviews. 
Screenshot 2024-04-14 at 3 05 59 PM
Screenshot 2024-04-14 at 2 57 44 PM

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,5 +29,8 @@
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^4.0.0"
+  },
+  "dependencies": {
+    "openai": "^4.33.1"
   }
 }
--- a/server/endpoints.ts
+++ b/server/endpoints.ts
@@ -15,7 +15,7 @@ export const configure = (app: Express) => {
     profileRouter,
     reviewRouter,
     courseRouter,
-    adminRouter,
-    aiRouter
+    adminRouter
   );
+  app.use('/ai', aiRouter);
 };
--- a/server/endpoints.ts
+++ b/server/endpoints.ts
@@ -5,6 +5,7 @@ import { profileRouter } from "./src/profile";
 import { reviewRouter } from "./src/review";
 import { courseRouter } from "./src/course";
 import { adminRouter } from "./src/admin";
+import aiRouter from "./src/ai/ai.router";
 
 export const configure = (app: Express) => {
   app.use(
@@ -15,5 +16,6 @@ export const configure = (app: Express) => {
     reviewRouter,
     courseRouter,
     adminRouter,
+    aiRouter
   );
 };
--- a/server/src/ai/ai.functions.ts
+++ b/server/src/ai/ai.functions.ts
@@ -1,0 +1,53 @@
+import dotenv from 'dotenv';
+import OpenAI from "openai";
+import { Reviews } from "../../db/schema";
+
+dotenv.config();
+const openai = new OpenAI({
+  apiKey: process.env.OPENAI_API_KEY,
+});
+
+/** Docstring for makeSummary. Takes in all reviews from a course as text and 
+ * creates a 50 word summary of those reviews.
+ * @params all reviews from a course
+ * @returns summary of reviews
+ */
+async function makeSummary(text: string) {
+  const completion = await openai.chat.completions.create({
+    model: "gpt-3.5-turbo",
+    messages: [
+      { role: "system", content: "You are creating a 50 word summary based on the collection of course reviews provided." },
+      { role: "user", content: text }
+    ],
+  });
+  return completion.choices[0].message.content;
+}
+
+/** Docstring for getCoursesWithMinReviews. We only want to create summaries for
+ * courses that have at least a certain number of reviews, so takes in that number
+ * and returns the courses that have at least that number of reviews.
+ * @params min reviews summary should be created for
+ * @returns courses ids with at least min reviews
+ */
+async function getCoursesWithMinReviews(minimum) {
+  const courses = await Reviews.aggregate([
+    {
+      $group: {
+        _id: "$class",
+        reviewCount: { $sum: 1 }
+      }
+    },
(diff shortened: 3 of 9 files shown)


server/src/ai/ai.functions.ts
```diff
@@ -0,0 +1,53 @@
...
+const openai = new OpenAI({
+  apiKey: process.env.OPENAI_API_KEY,
+});
+
+/** Docstring for makeSummary. Takes in all reviews from a course as text and 
+ * creates a 50 word summary of those reviews.
+ * @params all reviews from a course
+ * @returns summary of reviews
+ */
+async function makeSummary(text: string) {
```
qiandrewj, line 15: A question I have here is does it make a difference for ChatGPT to read a whole block of text, where reviews are strung one after the other, compared to us telling them this is a list of individual reviews?

server/src/course/course.controller.ts
```diff
@@ -65,15 +65,22 @@ export const getReviewsCrossListOR = async ({
...
-      copy.user = '';
-      return copy;
-    });
 
-    return sanitizedReviews;
+    // Return only the text from each review if returnTextOnly is true
+    if (returnTextOnly) {
+      return reviews.map(review => review.text).join(" ");
+    }
+    else {
```
qiandrewj, line 74: In the case where returnTextOnly is false, would returning sanitizedReviews be a safe way to handle the data because it could lead to errors with the OpenAI call?

qiandrewj (commented):
Helen - this work with the OpenAI API is really cool, and it'll be really interesting to keep developing with it and learning. From your test examples you showed, it looks like the AI call is working well to synthesize the key points from a larger amount of reviews, and it'll work well for the tags we want to develop later. Just left some questions about the functions interacting with the API that I was unsure about. Great work :)

package.json
```diff
@@ -29,5 +29,8 @@
     "eslint-plugin-jsx-a11y": "^6.2.3",
     "eslint-plugin-react": "^7.19.0",
     "eslint-plugin-react-hooks": "^4.0.0"
+  },
+  "dependencies": {
```
wizhaaa, line 33: ERMM we should probably delete this - 
I think it was when you did `yarn add openai -W` in the root and not just `/server` 

careful - be sure to not run yarn with this but tbh idk should probably be fine

server/endpoints.ts
```diff
@@ -5,6 +5,7 @@ import { profileRouter } from "./src/profile";
 import { reviewRouter } from "./src/review";
 import { courseRouter } from "./src/course";
 import { adminRouter } from "./src/admin";
+import aiRouter from "./src/ai/ai.router";
 
 export const configure = (app: Express) => {
   app.use(
```
wizhaaa, line 11: one thing we can do is 

`app.use('/ai', aiRouter)`

so we prefix `/ai` in front of all our ai endpoints

server/src/ai/ai.router.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const aiRouter = express.Router();
+
+aiRouter.use(express.json());
+
+/** Reachable at POST /api/summarizeReviews
+ * @body a block of text containing all reviews from a course
+ * returns a summary created by OpenAI
+*/
+aiRouter.post('/summarizeReviews', async (req, res) => {
```
wizhaaa, line 12: `/reviews/summarize/`

if we do the api.use('ai', ..) as before then this endpoint is reachable at `api/ai/reviews/summarize`

server/src/ai/ai.router.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/** Reachable at POST /api/getCourseIdsForSummary
+ * @body minimum number of reviews needed to create a summary
+ * returns all course ids that have at least that number of reviews
+*/
+aiRouter.post('/getCourseIdsForSummary', async (req, res) => {
+  try {
+    const min = req.body.min;
+    const ids = await getCoursesWithMinReviews(min);
+    if (ids === null) {
```
wizhaaa, line 33: Should we check against `null` or an empty list `[]` or if the length is 0?

server/src/ai/ai.router.ts
```diff
@@ -0,0 +1,50 @@
...
+      return res.status(400).json({
+        error: `No reviews found`,
```
wizhaaa, line 35: Would it be more accurate to say `no courses found with given minimum number of reviews`

server/src/ai/ai.router.ts
```diff
@@ -0,0 +1,50 @@
...
+      });
+    }
+
+    return res.status(200).json({
+      message: 'Retrieved all courses',
+      result: ids,
```
wizhaaa, line 41: tbh I forget how json response works, but can `result` be renamed `ids: ids` so its easier to know exactly what it returns

wizhaaa (commented):
Great work so far Helen! This feature is definitely coming along together very nicely.  

We discussed some parts about this feature, I'll leave some comments just so you can reference and give some pointers about optimization. 

ALSO yeah, upgrading to node 18 is quite a hassle D :  but I will try my best 

1. In the end, there should only be 1 endpoint ie `/api/ai/summarize` that does all these ednpoints and functionalities you have separated in one, but I like that it's separated now so I can tell how it is all working together - there are a lot of moving parts for sure. 

I'm not too sure about the endpoint in courses with the `returnTextOnly` boolean -> I think should make your own endpont / function eventually . but this is a fine workaround for now? 

Keep up the amazing work Helen! There are some blockers I didnt' really foresee but I'll try to see what we can do for now .

PR #442, opened 2024-04-14
Jacqueline/search algo

jacquelinecai:
# Summary

This PR works on the search bar/algorithm feature.

## PR Type

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [x] 🧑‍💻 Code Refactor
- [x] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Fixing case-sensitive bug:

image

Searching by full subject name:

image

## QA - Test Plan
- running the `Search.test.ts` file

## Breaking Changes & Notes
- fixed case-sensitive bug
- combined all search endpoints into a centralized endpoint
- updated test cases and frontend routes to reflect endpoint change
- modified edit distance search algorithm

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 notion
- [ ] 🍕 ...
- [ ] 📕 ...

[gif](https://media1.giphy.com/media/26n6WywJyh39n1pBu/200w.gif?cid=6c09b9522vo3bsf1yb1kwbndkvai9p7w9r9hiur95lffkqe4&ep=v1_gifs_search&rid=200w.gif&ct=g)


client/src/modules/SearchBar/Components/SearchBar.tsx
```diff
@@ -37,50 +37,33 @@ export const SearchBar = ({
...
-              setCourses(queryCourseList)
-            } else {
-              setCourses([])
-            }
-          })
-          .catch((e) => {
-            setCourses([])
-            console.log('Getting courses failed!')
-          })
+            const subjectList = response.data.result.subjects
```
wizhaaa, line 42: Will also say - _ideally_, we want the our endpoint to return only the data we want to work with. 

Further down the code, we slice each of these lists, ie. (0,3) and (0,5). The endpoint should only give us 3 items in subjects/professors and 5 in courses I think? 

IE. if we search 'cs' => I checked how big the course list length is (200) which is 195 data points we don't need the frontend to ever be exposed to => longer loading times and slower rendering (although these objects aren't that big + shouldn't make a big diff , it's still best practice.)

client/src/modules/SearchBar/Components/SearchBar.tsx
wizhaaa, line 1: Other comments pertaining to this file (I know you are not responsible for these, but these should be worked on): 

1. Key down function: on ArrowDown, there is no guard for going outside the indexable range. 
2. placeholder text is hardcoded based on initial page width and does not change on resize of window. (can potentially fix with a useEffect with screen width as the dependency)
3. In updateQuery function, should rename the let query to be something as let userInput so we do not mix it up with our query state variable. So like 
`let userInput = e.... => massage userInput => finally setQuery(userInput)`


renderResults is a messy function 
 - best practice is else return null instead of returning a div or just don't return anything at all 
 - should not be a function but a component ie. call it Results and then instead of calling the function as renderResults(), we render a component <SearchResults />  or <Results /> like so 
 - but naming might be tricky if you want to be smart because we have a page Results and a "results" list/dropdown type thing here 
 - instead of returning results list of "jsx elements", just return a component itself like
```javascript
<div> 
 <ProfessorsList />
 <SubjectsList /> 
 <CourseList /> 
</div>
```
where instead of 

```javascript
      professorList = professors.slice(0, 3).map((professor, i) => (
        <ProfessorResult
          ... 
      ))
```

we do ... 

```javascript
const ProfessorsResults/List/WhateverName = () => {
 return <div>
      { professors.slice(0, 3).map((professor, i) => (
        <ProfessorResult
          ... 
      )) } 
</div> 
} 
```

wizhaaa (commented):
Hi Jacqueline - really great work you have done so far relating to the search and its functionality. 

I tested this on my end and I think that main issue of the basic queries that should be working on mobile (due to case sensitivity on the front-end) has been fixed. 

I'm happy with the current implementation of this fix as well. 

It also looks like the test suite is acting up - let me know if you can fix it (or know the reason behind why there are errors (I think I see a timeout? - not sure myself -), else I will try to look into myself as well later this week (hopefully).

I can approve of this if you wish after test cases are tidied up. 

I do have quite a few comments regarding the search files and the algorithm itself - which can be left to a separate PR if you think that would be better/easier. 

Most of it is not your fault, but issues that have plagued the search files before either of us were on the team. We can also go through comments these during work session as well and evaluate from there. 



One thing my comments do not mention is kind of just how our search algorithm is built and how the main functionality of our site depends on searching. 

For example, the only way to find a course is through the search bar and if students are trying to find a course but our search algorithm is fully robust, it blocks them completely. 

I think our search algo works perfectly if students simply look up course subject + number, which if they are browsing with course roster/scheduler opened up, they can easily search up by sub + number. 

However, if they prefer to look up the name instead, our algorithm is not fully robust for that. (From my testing)

For example, I search up FWS, I see a class called Queer Cinema. 
Now if I search up Queer Cinema, nothing pops up. (at least nothing is rendered on the frontend). 

Personally, feels like a pretty big bug - we were able to find a course with one query but literally putting the course name ends up with no results. Same thing with 'object-oriented' or 'object oriented'. 

This would definitely require a pretty big overhaul of the searching system in general. This too can be worked on in a separate PR as well, and if you want to take on this challenge or not.

PR #443, opened 2024-04-14
Admin page refresh, admin manager modal

qiandrewj:
# Summary

This PR works on the Admin page feature, fixing old bugs and adding the ability to manage admin users by net-id.

## PR Type

- [X] 🍕 Feature
- [X] 🐛 Bug Fix
- [X] 📝 Documentation Update
- [X] 🎨 Style
- [X] 🧑‍💻 Code Refactor
- [X] 🔥 Performance Improvements
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Screenshot 2024-05-05 195112
Screenshot 2024-05-05 195151
## QA - Test Plan
- Tested all endpoints through rendering on front-end.
- Tested admin management modal by removing Jacqueline's admin privilege. She was then unable to access admin features. After re-adding her as an admin user, she then had access to the admin site.
## Breaking Changes & Notes
- Moved stats rendering (counting reviews, writing to CSV file) to backend. Split up endpoints in order to improve performance.
- Minor styling changes to whole Admin page.
- Updated old class components to functional components.
- Developed admin privilege interface.

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 notion
- [X] 📕 Method specifications
- [ ] 🙅 no documentation needed


leihelen (dismissed):
The admin page looks a lot better now and the changes you made definitely make everything easier to read and use. I think everything looks really good, however maybe a small visual change you could make would be to wrap the "admin interface", "pending reviews", and "reported reviews" labels in a css class as well and add some more margins or padding to align it with everything else. However, I'm unable to run the actual CU Reviews now on my local server and I'm not too sure why (I didn't meant to approve this request sorry).

leihelen:
The admin page looks a lot better now and the changes you made definitely make everything easier to read and use. I think everything looks really good, however maybe a small visual change you could make would be to wrap the "admin interface", "pending reviews", and "reported reviews" labels in a css class as well and add some more margins or padding to align it with everything else.

client/src/modules/Admin/Components/Stats.tsx
```diff
@@ -73,56 +68,33 @@ export default class Stats extends Component<Props, State> {
...
-        token: this.props.token,
-      })
-      .then((res) => {
-        const total = res.data.result
-        this.setState({ totalReviews: total })
-      })
-      .catch((err) => {
-        console.log('error retrieving totalReviews ', err)
-      })
+  getReviewsPerClassCSV() {
```
leihelen, line 91: Also I think for the CSV, since className shows the database id of the class, maybe you can make another endpoint after calling /fetchAllReviews that will get the class name from the classes database using the class id stored in each review and then map that to the review count instead.

leihelen (commented):
The admin page looks a lot better now and the changes you made definitely make everything easier to read and use. I think everything looks really good, however maybe a small visual change you could make would be to wrap the "admin interface", "pending reviews", and "reported reviews" labels in a css class as well and add some more margins or padding to align it with everything else. However, I'm unable to run the actual CU Reviews now on my local server and I'm not too sure why. This is the error I'm getting. 
PNG image

server/src/admin/admin.router.ts
wizhaaa, line 1: I think we should keep an endpoint for getting pending reviews and reported reviews instead of one endpoint to request a large number of review items (causing a heavy initial load time). 

IE. endpoints 
 - /api/admin/get/reviews/pending
 - /api/admin/get/reviews/reported 
 
Now, we do want "all" reviews to "count" how many each class has for our CSV. 
This should be handled in the backend and our frontend shouldn't handle any of this or need to load the review objects themselves to massage/do calculations with. 

For example, we can instead have an endpoint like: 
`/api/admin/get/reviewcountbycourse` or `/api/admin/get/csv/reviewcount`
 - this endpoint can either return a JSON of courseID/courseName (key) -> count (value) dictionary or just a `csv` string 
 - it should NOT return all review objects to our frontend => our frontend SHOULD NOT loop through all review objects and then generate on the frontend.

server/src/admin/admin.data-access.ts
wizhaaa, line 1: Should keep separate pending & reported reviews functionality

server/src/admin/admin.controller.ts
wizhaaa, line 1: Should keep separate pending & reported reviews functionality

client/src/modules/Admin/Styles/RaffleWinner.module.css
wizhaaa, line 1: Honestly, let's delete all the raffle winner code + style + functionality. But, do create a document on notion that summarizes what it does (ie. how it selects random students).

client/src/modules/Admin/Styles/Admin.module.css
```diff
@@ -0,0 +1,30 @@
...
+    flex: 1;
+}
+
+.semesterButtons {
+    background-color: var(--clr-gray-300);
+    color: var(--clr-white);
+    font-weight: var(--semi-bold-weight);
+    font-size: var(--font-small-size);
+    border-radius: 5px;
+    margin-top: 5px;
```
wizhaaa, line 15: 1. looks like this class is for a single button, so should rename to `semesterbutton` (singular) for best practice 
2. instead of uneven margins, consider wrapping all the buttons in a flex display with a set gap attribute 
ie: 
```
.buttonsContainer { 
display: flex;
flex-flow: row wrap;
gap: 5px;
... 

and then give each button a max width as well so it won't take the full row 
}
```

client/src/modules/Admin/Components/Stats.tsx
wizhaaa, line 1: Is this the old stats component? Consider just deleting the file and then making it from scratch - might be easier. 

It looks like this uses very old react syntax (class components) where you reference the component using `this`. You should update this to a functional component like: 

```typescript
const Stats = ( props: Props ) => { 

const { x, y, z } = props 
const [ state, setState] = useState() 

return <div> </div> 
```

Especially since I think a lot of this code (ie get chart data) is obsolete and more of a hassle to try to fix than build again. => should also delete any code that makes up the chart functionality (CSS, server-side code like endpints, functions, etc.)

client/src/modules/Admin/Components/AdminReview.tsx
wizhaaa, line 1: Also change this to functional component (not class)

client/src/modules/Admin/Components/Admin.tsx
```diff
@@ -49,20 +52,26 @@ export const Admin = () => {
...
+          setPendingReviews(
+            result.filter((review: Review) => review.reported === 0 && review.visible === 0)
           )
           setReportedReviews(
-            result.filter((review: Review) => review.reported === 1)
+            result.filter((review: Review) => review.reported === 1 && review.visible === 0)
           )
         } else {
-          console.log('Error at fetchPendingReviews')
+          console.log('Error at fetchAllReviews')
```
wizhaaa, line 74: Looks like there are console logs initially - (from before you worked on it) - fine for testing right now but should delete all console logs when you are done

client/src/modules/Admin/Components/Admin.tsx
wizhaaa, line 1: We discussed a lot of this already - but yep - separate the reviews endpoint into reported & pending and do not return all reviews to the frontend! 

A quick solution I can think of right now is: 

```typescript

useEffect( 

axios.hit('api/admin/get/reviews/pending' and '.../reported`) 
=> setLoading(false) => setPendingReviews( res.data.pending ... ) , setReported( res.data.reported ... ) 

)
```

``` javascript

return <div> ... 
<PendingReviews props = { ... } /> 
<ReportedReviews />  // might want to create a new component to make the Admin page less messy and chaotic 
</div>

```

wizhaaa (commented):
Great work Andrew! This is a very heavy task - a lot of sifting through old code and trying to figure how it all works. I also don't think I was super helpful either giving you pretty ambiguous directions and goals. After looking through your code and the admin page a bit more, I'll try to consolidate what I would like to seen accomplished (hopefully by the end of the semester). Do as much as you can for now - and don't worry too much if you can't get it all done. 

 - [x] Fix the loading time by separating endpoints (discussed) 
 - [ ] Update old class components to functional components 
 - [ ] Remove any outdated and unused code (unused states, functions, endpoints) such as raffle winner and stat charts (let me know if there is anything else you are considering) 
 - [ ] CSV generation done in back end 
 - [ ] Admin & user privilege management tool/interface (will try to sketch this one out) - feel free to make a sketch of your initial idea and design too! and we'll give feedback or iterate upon it 
 - [ ] actual updating roster buttons & functionality - [ a hard one, first, if you can summarize and breakdown how they work, that would be a helpful first step ( because this is kind of the back bone of our database and app functionality :o  ) ] 

image
 - should add a wrapper for the admin page to set a max-wdith (ie. 1440px is standard) 
 - add some left-right and top-bottom padding too (ie `padding: 64px 32px`), we can play around with this too and see what looks best. 

I left some other comments, a lot of which we already discussed but I wanted to put it in the PR. You probably updated a lot since then but haven't pushed yet - so some comments might already be addressed. Keep up the great work! This tasks is definitely very complex, requires quite a bit of thought, lots of moving pieces, and low-key pretty impactful.

PR #445, opened 2024-05-01
Attempt to upgrade to node 18 

wizhaaa:
# Summary

This PR works to upgrade to the next major node version - 18. 
Node 18 introduces a lot of security improvements and many libraries we want to use requires node 18 (ie. OpenAI)... 

Also does not run the test script - replaced with a no-op command `:` as I think Jest will break too much and isn't worth it right now. 

Be sure to update your global default node if you don't want to manually type out `nvm use 18` every time! 
->
```
nvm alias default 18
```

## PR Type

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 🧑‍💻 Code Refactor
- [x] 🔥 Performance Improvements
- [x] ✅ Test
- [x] 🤖 Build
- [x] 🔁 CI

## Breaking Changes & Notes

- change ur default node version to use node 18 
```
nvm alias default 18
```
- server explicitly uses typescript 5.4 in dependencies 
- react-scripts SSL 

Using the SSL flag for our react-script functions is not actually ideal - as it definitely introduces some security concerns - but I think that this is for the best. 
IE. react scripts only really supports node 16. This is the main workaround I could find for react-scripts. The long term solution would be to change from react-scripts to vite (which I attempted earlier but required node 18 - but may attempt once again.) 

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 notion
- [ ] 🍕 ...
- [ ] 📕 ...
- [ ] 🙅 no documentation needed


qiandrewj (approved):
Hi Will, the upgrade to node 18 works for me. I think that the considerations for vite over react-scripts and overhauling the test script are important to work on, but we can first make the move to node 18 so that it's easier to develop those changes :)

PR #457, opened 2024-09-22
generate tags

leihelen:
# Summary

This PR works on generating tags for the Cornellians Say feature.

## PR Type

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

Screenshot 2024-09-22 at 12 09 42 AM
Tags generated for info 2300 in order of lectures, assignments, professor, skills, and resources. First element for each tag is the adjective corresponding to which tag it is for and the second element for each tag is the connotation of that adjective. 

## QA - Test Plan
Tested different prompts and tried different kinds of tags on multiple different courses. 

## Breaking Changes & Notes
Uses all reviews from a course to generate an array which can be used to create the tags on the Cornellians Say summary and also color the tags red, yellow, or green corresponding to a negative, neutral, or positive tag. 

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 notion
- [x] 🍕 docstrings


server/src/ai/ai.functions.ts
```diff
@@ -40,6 +40,50 @@ async function makeSummary(text: string) {
...
+ * @returns an array of adjectives and their corresponding connotations describing
+ * lectures, assignments, professr, skills, and resources in that order
+ */
+async function generateTags(text: string) {
+  const completion = await openai.chat.completions.create({
+    model: "gpt-3.5-turbo",
+    messages: [
+      {
+        role: "system", content: `
+          You are creating 5 adjectives describing the lectures, assignments, professor, skills, and resources,
```
wizhaaa, line 58: Should we tell it that it is given a string of consecutive reviews for a single course?

server/src/ai/ai.functions.ts
```diff
@@ -40,6 +40,50 @@ async function makeSummary(text: string) {
...
+      { role: "user", content: text }
+    ],
+  });
+  const response = completion.choices[0].message.content;
+  const tagsArray = response.split(',').map(item => {
+    const match = item.match(/: (.+) \((.+)\)/);
+    if (match) {
+      const adjective = match[1].trim();
+      const connotation = match[2].trim();
+      return [adjective, connotation];
```
wizhaaa, line 78: Should we also include the actual category in our response so that we don't have to remember the order ourselves? 

Potential idea is to create an array of the categories and loop through it to add like so: 

```typescript
const categories = ["Lectures", "Assignmnets", ... ]; 
categories.forEach( (category) => { 
// ...
return [category, adjective, connotation]; 
}); 
```

server/src/ai/ai.router.ts
```diff
@@ -23,6 +23,27 @@ aiRouter.post('/text/summary', async (req, res) => {
   }
 });
 
+/** Reachable at POST /api/ai/text/tags
+ * @body a block of text containing all reviews from a course
+ * returns an array of tags along with their connotation created by OpenAI
+*/
+aiRouter.post('/text/tags', async (req, res) => {
+  try {
+    if (!req.body.text) {
```
wizhaaa, line 32: Since we use `text` multiple times, it might be easier to decompose it like so:

```typescript
const { text } = req.body;
```

wizhaaa (approved):
Awesome work Helen using OpenAI to generate tags for course summaries! 

Few questions (we can go over in-person as well):
1. Is this PR mainly to test out the tags feature? 
2. Do you want separate API endpoints for the final version for tags and the summary or just one endpoint that calls OpenAI once and asks for the summary + tags at the same time? 

Great work so far tho - love to see the tag generation working as well!

server/src/ai/ai.router.ts
```diff
@@ -1,28 +1,33 @@
...
     console.error(err);
     res.status(500).json({ error: 'Internal Server Error' });
   }
 });
 
+/** Reachable at POST /api/ai/get/text
+ * @body a course ID
+ * returns a block of text containing all reviews from that course
+*/
 aiRouter.post('/get/text', async (req, res) => {
```
wizhaaa, line 31: also maybe change this to 
`api/ai/get-course-reviews` or `api/ai/get-course-reviews-text`?

server/src/ai/ai.router.ts
```diff
@@ -1,28 +1,33 @@
...
 aiRouter.use(express.json());
 
-/** Reachable at POST /api/ai/summarizeReviews
+/** Reachable at POST /api/ai/text/tags
  * @body a block of text containing all reviews from a course
- * returns a summary created by OpenAI
+ * returns a dictionary containing the summary and tags created by OpenAI
 */
-aiRouter.post('/text/summary', async (req, res) => {
+aiRouter.post('/text/summary/tags', async (req, res) => {
```
wizhaaa, line 13: what about `/api/ai/summarize-reviews`?

wizhaaa (approved):
Really good work on this PR! I like how you returned a dictionary of the summary and the tags all in one endpoint and one OpenAI API call. 

I left some comments about the endpoint names but these can always be changed in the future. But it might be helpful to create a document that summarizes this feature and all related endpoints and its main functionalities to help see naming cohesion. 

Also, I was thinking that you might want to return the tags as a dictionary as well and the keys are the tag names. So when you do work on the frontend you could access the values by like `tags.lectures.adjective` or `tags.lectures[0]`. Storing it in a dictionary just helps debugging in the frontend and makes more clear what you are working with. 

But, an array works too because you can just loop over it. Up to you since you will be implementing the front end

PR #463, opened 2024-10-14
Initialize new database fields

leihelen:
# Summary

This PR works on the Cornellians Say feature.

## PR Type

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

## QA - Test Plan
-Leave a new review and then check that the freshness count for that class has gone up accordingly. 
-Check that summary, tags, and freshness are updated accordingly in the database when calling the ai function.

## Breaking Changes & Notes
-new fields added to the Classes schema to hold the ai generated summaries and tags (optional fields that will be filled for classes with at least 3 reviews) as well as a freshness count
-adding new reviews should update the freshness count by one
-added function that will generate summary and tags and reset freshness for a given course

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 notion
- [x] 🍕 docstrings


server/src/review/review.data-access.ts
```diff
@@ -49,6 +49,11 @@ export const findClassReviews = async (courseId: string) => {
 export const insertReview = async (review: Review) => {
   const newReview = new Reviews(review);
   await newReview.save();
+  const courseId = review['class'];
+  await Classes.updateOne(
+    { _id: courseId },
+    { $inc: { freshness: 1 } }
```
qiandrewj, line 55: A couple thoughts on updating freshness: first, I believe the scale that you and Will were discussing was out of 100. Considering that we don't get a ton of reviews, we could consider a couple different ideas to better work within the site:

- Increase freshness by more (i.e. 5) each review
- Perhaps by a constant factor (increase by 10%) of the current freshness. This might represent that a lot of ongoing activity exponentially increases the freshness (so it increases quickly?), but could also foresee issues with this and the freshness limit

Those would probably depend on how quickly freshness depreciates. We'd also want to have a cap on freshness so we cannot increment the field past its max limit.

qiandrewj (commented):
Hi Helen! The updates to the database look good to me, otherwise on the "freshness" update I feel like we could look into it a bit more so that freshness scales correctly when either a person leaves a review or it decays naturally over time. Great work :)

server/src/review/review.data-access.ts
wizhaaa: Possible options of the top of my head:

1. Current implementation (increment freshness by one per new review) 
-> when we run updateSummaries in a new semester, we have to pull freshness and divide by number of reviews. => have some threshold here (ie.

```python
if freshness/number of reviews >= 30%: 
   resummarize()
else: 
   continue
```


2. We pre-calculate freshness upon insert

wizhaaa (approved):
Everything looks pretty solid so far - added a few comments on endpoint names and freshness. Just have to fix the merge conflicts - can help if u need. 

I would also think it would be helpful if you created a document or something that has all the endpoints + main functions + req body + response types 


IE it can look like: 

POST `/api/ai/summarize-all-courses` 
admin token => bool 
Triggered by a button in frontend (Admin page) -> populates our database with new review summaries for courses. 

POST `api/ai/get-course-review-text` 
courseID -> text 
Takes a courseId and returns all the reviews for a course squished into one string (separated by `/`) 

etc... and maybe a diagram that shows how they interact with each other would be helpful

PR #480, opened 2024-11-30
Course panels and gauges redesign

qiandrewj:
# Summary

This PR works on the course left/right split and gauges redesign feature.

## PR Type

- [X] 🍕 Feature
- [X] 🎨 Style

## QA - Test Plan
Tested on course page and search results page, both for courses with data and courses with no reviews (so ratings are undefined). Responsiveness tested as well.

## Breaking Changes & Notes
New gauges component with new styling replaces old gauge component. Replaced in both the courses page and results page.

## Added to documentation?

- [X] 🙅 no documentation needed


jacquelinecai (approved):
Thank you for creating the panels and implementing the new gauges UI! I left a couple comments based on my frontend testing and code style, but overall it looks great!

client/src/modules/Course/Components/Gauges.tsx
```diff
@@ -0,0 +1,189 @@
...
+            setDifficultyEmote(undefined)
+            setDifficultyBars(['0%', '0%', '0%', '0%', '0%'])
+            setDifficultyHover('')
+        }
+    }, [difficulty])
+
+    useEffect(() => {
+        // Higher workload scores are more negative
+        if (workload) {
+            if (0 <= workload && workload < 3) {
```
leihelen, line 107: Maybe some of the difficulty and workload code could be combined into one function to avoid repetitive code.

leihelen (approved):
Everything looks great on my end too, good job Andrew! Something I noticed that could maybe be improved in the future is just that the content on the right sidebar in the page before leaving a review begins a little lower than the left side. Also another idea is maybe the left and right sidebars on both pages could be more consistent in size so its a little smoother when navigating from page to page. Overall, everything looks good and this will be very helpful for implementing the frontend for similar courses and cornellians say!

PR #492, opened 2024-12-16
fix capitalization issue

leihelen:
make sure all adjectives on tags are uppercase


jacquelinecai (approved):
Just a quick comment!

PR #494, opened 2024-12-24
Filter by prof on course page

mjaydenkim:
# Summary

This PR works on the filter by professor course page dropdown.

## PR Type

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

https://github.com/user-attachments/assets/702f3af8-4f84-4440-ae53-e0308b45ea3a
image

## QA - Test Plan
Checked on different screen sizes, fixing styles as needed. Looks slightly awkward on mobile since the header brushes up against the text, but they never overlap. Checked on a variety of classes to ensure that all professors were added to the dropdown. 

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 notion
- [ ] 🍕 ...
- [ ] 📕 ...

[gif](https://giphy.com/gifs/talkshowgameshow-episode-19-talk-show-the-game-26DN0TFE3TXYJq70A)

### DIFF
--- a/client/src/modules/Course/Components/Course.tsx
+++ b/client/src/modules/Course/Components/Course.tsx
@@ -315,42 +315,40 @@ export const Course = () => {
             <div className={styles.reviewscontainer}>
               <div className={styles.bar}>
                 <h2 className={styles.title}>
-                  Past Reviews (
-                  {selectedProf.current !== 'none'
-                    ? visibleCourseReviews.length + ' of '
-                    : ''}
-                  {courseReviews.length}){' '}
+                  Past Reviews ({visibleCourseReviews.length}){' '}
                 </h2>
                 <div>
                   <div className={styles['select-container']}>
-                    <label htmlFor="sort-reviews">Sort by: </label>
-                    <select
-                      name="sort-reviews"
-                      id="sort-reviews"
-                      onChange={sortReviewsBy}
-                      className={styles.filtertext}
-                    >
-                      <option value="helpful">Most Helpful</option>
-                      <option value="recent">Recent</option>
-                      {selectedProf.current === "none" && <option value="professor">Professor Name</option>}
-                    </select>
-                  </div>
-                  <div className={styles['select-container']}>
-                    <label htmlFor="filter-by-prof">
-                      Filter by professor:{' '}
-                    </label>
-                    <select
-                      name="filter-by-prof"
-                      id="filter-by-prof"
-                      onChange={filterByProf}
-                      className={styles.filtertext}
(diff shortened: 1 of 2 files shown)


client/src/modules/Course/Styles/Course.module.css
```diff
@@ -167,6 +168,13 @@
   color: var(--clr-gray-300);
 }
 
+.select-container {
```
qiandrewj, line 171: Would love if both sorts/filters could be put into a flexbox, also so that they would both take up the same amount of space.
image
It would look best if the filter and sort both ended at the same distance, so it looked more rectangular. By encapsulating them in a div together, it would also be helpful for our future design (which I think is to move the sort and filters to below the title)

client/src/modules/Course/Components/Course.tsx
```diff
@@ -279,52 +310,73 @@ export const Course = () => {
...
-            </div> */}
           </div>
           <div className={styles.rightPanel}>
-            {/* Reviews Displaying */}
             <div className={styles.reviewscontainer}>
               <div className={styles.bar}>
-                <h2 className={styles.title}>Past Reviews ({courseReviews?.length}) </h2>
+                <h2 className={styles.title}>
+                  Past Reviews (
+                  {selectedProf.current !== 'none'
```
qiandrewj, line 319: It's more intuitive for me to have it as just `visibleCourseReviews.length` because otherwise, you might interpret it as being on a second "page" of reviews or something. Can have others give feedback too

qiandrewj (commented):
This works super cleanly, Matt! It's great now that we have multiple sorting/filtering methods that work seamlessly with each other. Left a few comments, but thanks :)

client/src/modules/Course/Components/Course.tsx
mjaydenkim: oh wait this makes a lot of sense!! my logic behind this was that i wanted a way to make sure that users easily know that when a professor is selected they're not seeing all the reviews (esp b/c the statistics on the left sidebar are based on all reviews). i think just changing the number itself works though!

qiandrewj (approved):
Thank you for addressing those comments, Matt - this will be super helpful!! Fixed a tiny styling thing that was bothering me

PR #495, opened 2024-12-24
Less intrusive anon modal

mjaydenkim:
# Summary

This PR works on streamlining the semi-intrusive anon modal.

## PR Type

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [x] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

image
image

## QA - Test Plan
Tested on a variety of sizes and with logging in/out repeatedly.

## Breaking Changes & Notes
Kept the modal around but changed its name from AnonymousWarning to LoginModal. Its only purpose is now to tell the user to log in after submitting a review without first having done so. All uses of the class should have been modified. I'm using Jaeyoung's adorable bear emoji for this class, if that's ok!

Now, there's text above the submit button informing users of their anonymity. Unlike the modal, this notice is always present, no matter how many reviews the user has written. 

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 notion
- [ ] 🍕 ...
- [ ] 📕 ...


client/src/modules/Course/Styles/ReviewModal.module.css
```diff
@@ -119,6 +119,16 @@
   place-self: center;
 }
 
+.anonymouslabel {
+  display: flex;
+  font-size: 1em;
```
qiandrewj, line 124: For consistency with the codebase, it'd be best to use pixels for margin size and font size

qiandrewj: I also think that since it's in a flex-column already, you might not need bottom and top margins because it should just use the gap attribute - you can play around with decreasing the gap size in `.textcol` since you added a new text row.

client/src/modules/Course/Components/LoginModal.tsx
```diff
@@ -0,0 +1,50 @@
...
+      <div className={styles.emoji}>
+        <img
+          src={Emoji}
+          className="emoji"
+          alt="CUReviews Bear -- 'You got this!' emoji"
+        />
+      </div>
+      <div className={styles.message}>
+        <span className="line1">
+          Login to submit - your review will upload once you login!
```
qiandrewj, line 32: Message here is a bit redundant - I think just "Your review will upload once you login!" is great

qiandrewj (commented):
This looks great, Matt! Left a few comments about styling and code style, but otherwise it worked well on my end :)

client/src/modules/Course/Components/ReviewModal.tsx
```diff
@@ -251,6 +232,9 @@ const ReviewModal = ({
               id="review-content"
               placeholder={placeholdertext}
             ></textarea>
+            <p className={styles.anonymouslabel}>
```
jacquelinecai, line 235: image
Is it possible to make this text centered between the text box and the submit button?

jacquelinecai (commented):
Great work Matthew! This looks a lot more intuitive on the user side.

One thing about the emoji used--could you add it to the codebase as an svg instead to maintain consistency? There should be examples of this within Cornellians Say and Similar Courses features already!

client/src/modules/Course/Components/ReviewModal.tsx
```diff
@@ -79,25 +77,8 @@ const ReviewModal = ({
...
-    const response = await axios.post('/api/profiles/count-reviews', {
-      netId
-    });
-    const res = response.data;
-    if (response.status === 200) {
-      setNoReviews(res.result === 0);
-    }
-  }
-
   function onProfessorChange(newSelectedProfessors: string[]) {
```
jacquelinecai, line 82: This can be part of a codebase cleanup PR, but it'd be nicer if these functions were all arrow functions (i.e. `const onProfessorChange = (newSelectedProfessors: string[]) => {...}`).

mjaydenkim: I only converted the four onXChange methods -- should I convert the other functions as well?

qiandrewj: Yeah, that would be great!

jacquelinecai (approved):
Thanks for all the changes, Matt. Looks great!

PR #501, opened 2025-01-17
Add major selection to profile

mjaydenkim:
# Summary

This PR works on a major update... haha i'm so funny...... anyway it allows users to set their major as a part of their account rather than each review individually (although reviews stay attached to the major selected at the time of publication). 

New users are prompted to select their major on the review page with the same UI as before, but their selection will then attach to their account. Users shouldn't be shown the major selection dropdown otherwise as long as they continue to have a selected major and are logged in. Existing users can update their major on the profile page, where there's a dropdown menu to do so. 

## PR Type

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 🎨 Style
- [ ] 🧑‍💻 Code Refactor
- [ ] ✅ Test
- [ ] 🤖 Build
- [ ] 🔁 CI
- [ ] 📦 Chore (Release)
- [ ] ⏩ Revert

https://github.com/user-attachments/assets/88ed0c9b-105f-4a37-a8fc-bf0b5bad2a08
https://github.com/user-attachments/assets/8c0ca687-3eb0-4a21-aa40-6be17db0dfa1
majorupdate_ss1

## QA - Test Plan
Tested in a variety of different account scenarios, including with accounts that have no major, accounts with many majors, etc. Tested new UIs on mobile, but there isn't a ton that's different. Couldn't test with a new account (b/c I don't have other NetIDs) but I simulated that performance essentially. Also tested to ensure that majors are still correctly tied to reviews.

## Breaking Changes & Notes
- All logged-out users are shown the major dropdown, and selecting a new major for an existing user notifies that user (via toast) and changes the major to the more recent selection. 

## Added to documentation?

- [ ] 📜 README.md
- [ ] 📓 notion
- [ ] 🍕 ...
- [ ] 📕 ...

### DIFF
--- a/client/src/modules/Profile/Component/UserInfo.tsx
+++ b/client/src/modules/Profile/Component/UserInfo.tsx
@@ -11,7 +11,6 @@ type UserInfoProps = {
   reviewsTotal: number;
   upvoteCount: number;
   netId: string;
-  majors: string[];
   signOut: () => void;
 }
 
--- a/client/src/modules/Profile/Component/UserInfo.tsx
+++ b/client/src/modules/Profile/Component/UserInfo.tsx
@@ -3,6 +3,7 @@ import styles from '../Styles/UserInfo.module.css';
 import ProfileCard from './ProfileCard';
 import MultiSelect from '../../Course/Components/MultiSelect';
 import allMajors from '../../Globals/majors';
+import formatList from 'common/formatList'
 import axios from 'axios';
 import { toast } from 'react-toastify';
 
@@ -68,9 +69,7 @@ const UserInfo = ({
           <span className={styles.bold}>{netId}</span>
           {userMajors.length > 0 && " is studying"}
           <p className={styles.bold}>
-            {userMajors.join(', ').replace(/(, )(?!.*\1)/,
-              (userMajors.length > 2 ? ', and ' : ' and ')
-            )}
+            {formatList(userMajors)}
           </p>
         </div>
       </div>
--- a/client/src/modules/Profile/Component/UserInfo.tsx
+++ b/client/src/modules/Profile/Component/UserInfo.tsx
@@ -1,12 +1,17 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import styles from '../Styles/UserInfo.module.css';
 import ProfileCard from './ProfileCard';
+import MultiSelect from '../../Course/Components/MultiSelect';
+import allMajors from '../../Globals/majors';
+import axios from 'axios';
+import { toast } from 'react-toastify';
 
 type UserInfoProps = {
   profilePicture: string;
   reviewsTotal: number;
   upvoteCount: number;
   netId: string;
+  majors: string[];
   signOut: () => void;
 }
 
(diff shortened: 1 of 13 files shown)


leihelen (approved):
This looks really good and will be a really important feature since a lot of people are turned away from posting reviews because there are so many steps to complete so this will help simplify the process! I tested the same scenarios you went through and it works well on my end as well. Maybe you could add some try catch statements when you call the api and make post requests in case of unexpected errors as well.

client/src/modules/Profile/Component/UserInfo.tsx
```diff
@@ -17,6 +22,40 @@ const UserInfo = ({
...
+  }
+
+  const getUserMajors = async () => {
+    const response = await axios.post('/api/profiles/get-majors', { netId })
+    if (response.status === 200) {
+      setUserMajors(response.data.majors)
+    }
+  }
+
+  const updateMajors = async () => {
```
qiandrewj, line 47: This still makes sense to stay in the UserInfo component. However, I do think it needs a redesign because it looks clunky right now as-is

client/src/modules/Profile/Component/UserInfo.tsx
```diff
@@ -17,6 +22,40 @@ const UserInfo = ({
...
+    const response = await axios.post('/api/profiles/set-majors', { netId, majors: selectedMajors })
+    console.log(response)
```
qiandrewj, line 49: Since this is user-facing, we don't want to have console.logs. In the admin page, we're a bit more lax with it because it's only for our team anyway.

client/src/modules/Profile/Component/UserInfo.tsx
```diff
@@ -25,7 +64,15 @@ const UserInfo = ({
         src={profilePicture}
         alt="user profile bear"
       />
-      <div className={styles.netid}>{netId}</div>
+      <div>
+        <div className={styles.netid}>
+          <span className={styles.bold}>{netId}</span>
```
qiandrewj, line 69: I really like this new feature - looks clean too!

client/src/modules/Profile/Component/UserInfo.tsx
qiandrewj: The toast and frontend update looks smooth tho, the main thing is the button is really large and possibly in an awkward place

server/src/review/review.controller.ts
```diff
@@ -11,7 +11,7 @@ import {
   updateStudentLikedReviews,
   updateStudentReviews,
   updateReviewLikes,
-  hideReportedReview
+  hideReportedReview, updateStudentMajors
```
qiandrewj, line 14: add a line break

server/src/profile/profile.router.ts
```diff
@@ -3,10 +3,11 @@ import express from 'express';
 import { Profile } from './profile';
 
 import {
+  getStudentMajors,
   getStudentReviewDocs,
-  getTotalLikesByNetId
+  getTotalLikesByNetId, setStudentMajors
```
qiandrewj, line 8: add a line break

qiandrewj (changes requested):
Hi Matt, thank you for taking the initiative to do this (much-needed) quality of life update. The backend functions well for me, the only areas where I have some concerns are in the front-end, so let's speak to some designers so that this can become an official feature for CUReviews!

PR #516, opened 2025-03-02
Organizing assets in codebase

qiandrewj:
# Summary

This PR works on the codebase cleanliness. Prior, assets existed in many different parts of the codebase and imports were used inconsistently. There were a ton of duplicates as well due to this disorganization. Hopefully now, all assets will be self-contained in frontend.

## PR Type

- [X] 🧑‍💻 Code Refactor

N/A

## QA - Test Plan
Tested throughout to make sure that imports to all assets are consistently working.

## Added to documentation?

- [X] 🙅 no documentation needed


mjaydenkim (approved):
this all looks incredible, tysm!! i was wondering what a flavicon is 😭😭

my only note (an impossibly small nitpick) is that the new report icon kinda looks clunky given that it's a filled-in icon when the upvote clap emoji next to it is an outline, it kinda clashes w/ the design language of the page. i don't think it's worth fixing though if it's not really easy, it literally doesn't matter

leihelen (approved):
This looks really good and definitely makes the code a lot cleaner to read! I've definitely struggled a bit with navigating through the assets in the past so this is a very helpful change! However, I think maybe it would make sense to keep the different backgrounds in the public asset folder still?

